Issue #95, opened 2025-06-06
Task: avbufferhelper: Update the interface


bhanucbp:
### Goal

The avbufferhelper interface definition does not align with the actual vDevice implementation.

https://github.com/rdkcentral/rdk-halif-aidl/blob/main/avbuffer/current/avbufferhelper.h

PR #99, opened 2025-06-12
added #95 baseline code

Ulrond:
Upgrade avbuffer interface after bring up of vDevice.


outdooruseonly (dismissed):
Needs testing that the lib implementation can inherit the base class and the needed pure virtual function.

Ulrond:
> ds testing that the lib implementation can inherit the base class and the needed pure virtual function.

Upgraded to pure virtual function

PR #109, opened 2025-06-17
Added 102 task c++ singleton av buffer helper

Ulrond:
Created a simple C++ Singleton using a simple interface fully public and defined.


Ulrond:
Related to -> https://github.com/rdkcentral/rdk-halif-aidl/pull/110/files

Ulrond:
served its purpose to showcase a solution and is no longer required.

PR #175, opened 2025-08-26
Feature/MRITE-122 extending broadcast hal

japsen73:


Ulrond:
- Is this the first commit even in rdkcentral?If not then your commit author & id information is incorrect for the creds registered in rdkcentral.
- Branch names are based off github issues, so that it's clear why from the release history things are changing, can you rename your branch and create a ticket on what your doing,, MRITE-122 is an internal reference not a github issue one. https://github.com/rdkcentral/rdk-halif-aidl/issues, to describe the who what why where, and the branch naming convention is required.
- not stuck on /broadcast as a single directory, we should separate out, consider /frontend, /lnb, /tuners or broadcast/frontend, broadcast/lnb, broadcast/tuner... Through if we do that, this would be different than all the other directories and may cause issues with the scripts,

Issue #204, opened 2025-09-26
Task: Deepsleep : HFP file needs modification


thanushreevani:
### Goal

https://github.com/rdkcentral/rdk-halif-aidl/blob/main/deepsleep/current/hfp-deepsleep.yaml

DeepSleep module name should be inline with service name and needs upgrading.

Notes (Optional)

Ulrond:
@thanushreevani - please review -> https://rdkcentral.github.io/rdk-halif-aidl/0.11.0/external_content/ut-core-wiki/1.2.-Standards%3A-Issue-Description-Guideline/, and be specific about what issue your seeing here?

Ulrond:
are you stating that this

```yaml
deep_sleep:                 # Component object begins
    interfaceVersion: current
```

is not aligned, please upgrade your description in the ticket to be specific?

PR #205, opened 2025-09-26
gh #204 Deepsleep HFP file update

thanushreevani:
Update module name to be inline with Servicename


deepsleep/current/hfp-deepsleep.yaml
```diff
@@ -22,7 +22,7 @@
 
 # HAL Feature Profile for deep_sleep
 
-deep_sleep:                 # Component object begins
+DeepSleep:                 # Component object begins
```
Ulrond, line 25: that's not correct, the interface name is `IDeepSleep`, it's not the `serviceName`

Issue #208, opened 2025-09-26
Task: Indicator : HFP file needs modification


KarthikeyanR470:
### Description:

**Problem:** Indicator module name should be inline with interface name and needs upgrade

**Steps to Reproduce:** https://github.com/rdkcentral/rdk-halif-aidl/blob/main/indicator/current/hfp-indicator.yaml 

**Expected Behaviour:** Root node should be `IIndicator`
**Actual Behaviour:** Root node is `indicator`

Ulrond:
you need to be more specific in the description to describe what the problem is your trying to resolve? Stuff todo isn't a good enough description.. @KarthikeyanR470

Ulrond:
@KarthikeyanNanjundan  please update the description before you proceed

PR #209, opened 2025-09-26
gh #208 Indicator HFP file update

KarthikeyanR470:
**Problem:** Indicator module name should be inline with interface name and needs upgrade

**Expected Behaviour:** Root node should be IIndicator
**Actual Behaviour:** Root node is indicator

### DIFF
--- a/indicator/current/hfp-indicator.yaml
+++ b/indicator/current/hfp-indicator.yaml
@@ -31,7 +31,7 @@
 # * the valid set of states for the indicator service on a given device.
 # *
 
-Indicator:
+IIndicator:
   interfaceVersion: current
   supportedStates:
     # General system states


indicator/current/hfp-indicator.yaml
```diff
@@ -31,7 +31,7 @@
 # * the valid set of states for the indicator service on a given device.
 # *
 
-indicator:
+Indicator:
```
Ulrond, line 34: name should be inline with the interface name of `IDeepSleep`, not the service name, 

the interface is being described here..

Ulrond (changes requested):

Ulrond (changes requested):
IIndicator is required.

Issue #201, opened 2025-09-24
create new parcelable within the FrameMetadata to handle the AC3 AC4 EAC3 Pass through mode


shafi12:
### Problem/Opportunity

Some audio decoders operate in pass-through mode, where the encoded audio data (e.g., AC3, EAC3, or AC4) is not decoded but instead forwarded directly to the next element in the pipeline. In such cases, it's essential to distinguish—via metadata—whether the decoder's output is raw encoded data or decoded PCM audio, so that the audio sink element can handle the stream appropriately.

### Proposed Solution

To create the  parcelable within the FrameMetadata to handle the AC3 AC4 EAC3 encoded data

PR #227, opened 2025-10-08
Feature/201 frame meta data tohandle private data

shafi12:
https://github.com/rdkcentral/rdk-halif-aidl/issues/201


Ulrond (changes requested):
the spacing used here, is not consistent with the rest of the file.

PR #237, opened 2025-10-14
Feature/gh213 add service name for flash

### LINKED ISSUE #213: Bug: serviceName is not available for Flash Module
### Problem/Opportunity

The service name for Flash Module is not available
getServiceName() is generated by linux binder for flash module

### Steps to reproduce

The service name for Flash Module is not available

### Expected Behavior

  const @utf8InCpp String serviceName should be added for flash

### Actual Behavior

serviceName not available for flash

### Notes (Optional)

_No response_

revram16:
**Pull request overview**
This PR standardizes the Flash HAL AIDL interface by adding an explicit published service name constant, aligning Flash with other HAL modules that expose a serviceName in their AIDL.

Changes:

- Add const @utf8InCpp String serviceName = "flash"; to IFlash.aidl.


Ulrond (changes requested):
missing approval from the original ticket which has missing information to move this forward.

Ulrond:
@kanjoe24  - can you re-write this one? @revram16

Ulrond (approved):

PR #239, opened 2025-10-14
added #234 client usage document

### LINKED ISSUE #234: Task: Documentation on how to use versions as a client
### Goal

Need to explain how the upper layers would support versioning control of functions and changes over time in the interface.

The upper layers are common but must adapt to features that would be missing on earlier servers, compared to latest ones.

### Notes (Optional)

Example of this problem.

V1 - functionA, V2 -> functionA + functionB

Ulrond:
Added support for documentation on the client usage of controlling versions on the interface.


Ulrond:
merged to develop after review

Issue #22, opened 2025-02-05
Task: Flash HAL & Review and check documentation


Ulrond:
### Subject

- Review based on latest feedback, rename to Firmware / Image? 
- Discuss with team, review current docs

- `IFlash` -> `IUpdater` -> `updateFirmware`?

### Notes (Optional)

- Flash HAL -> https://ccp.sys.comcast.net/browse/CPESP-6021

Ulrond:
Closing as superseded — Flash HAL documentation work landed in 0.15.0 via PRs #22, #340, and #370 ("docs: Flash HAL review documentation"). The Flash HAL interface itself is also slated for **renaming** under a follow-up issue (the term "flash" isn't the right name for this HAL's actual scope of firmware image management). Tracking that separately.

PR #241, opened 2025-10-15
Feature/gh236 remove error8 flash image result

revram16:
Removed ERROR_FLASH_VERIFY_SIGNATURE_FAILED = 8, from image result


Ulrond:
const @utf8InCpp String serviceName = "Flash"; -> Will have to be covered under a different ticket #22

Issue #291, opened 2025-12-05
Bug: Panel: Include VideoDecoder interface file not added in CMakeLists.txt


KarthikeyanR470:
### Problem/Opportunity

In Panel module, VideoDecoder src file and its path not added in CMakeLists.txt
videodecoder/current/com/rdk/hal/DynamicRange.aidl

Causing Build error

### Steps to reproduce

clone repo for Panel
./build.sh Target=linux

### Expected Behavior

Build should be success

### Actual Behavior

Build fails with no such file
CMakeLists.txt not pointing to VideoDecoder src file path.

PR #292, opened 2025-12-05
gh #291 added videodecoder src file in cmakelist.txt

KarthikeyanR470:
In Panel module, VideoDecoder src file and its path not added in CMakeLists.txt
videodecoder/current/com/rdk/hal/DynamicRange.aidl

Causing Build error


Ulrond:
@KarthikeyanR470 -is this still required?

PR #324, opened 2026-01-16
baseline #323 0 Right format document

### LINKED ISSUE #323: Task:White Paper, Right Format for the Right Job
### Goal

Need to create a whitepaper on json vs yaml vs toml vs ini..

So we can guide when to use what why where

### Notes (Optional)

_No response_

Ulrond:
Created new doc to define a model for which format should be used for what


kanjoe24 (approved):
Approved.

PR #408, opened 2026-04-10
task:compositeinput: generation 2 — PortProperty enum, port lifecycle, telemetry consolidation (#382)

### LINKED ISSUE #382: Gaps in CompositeInput AIDL Interfaces
### Discussed in https://github.com/rdkcentral/rdk-halif-aidl/discussions/379

<div type='discussions-op-text'>

<sup>Originally posted by **thanushreevani** March 30, 2026</sup>

1. HFP - no info available for the PlatformCapabilities. Only port specific capability parameters available.
           https://github.com/rdkcentral/rdk-halif-aidl/blob/8ab4217b36a2a614384f5276343fc67226a3a236/compositeinput/current/com/rdk/hal/compositeinput/PlatformCapabilities.aidl#L31
2. HFP - There is no seperate enum/list for the list of property types available.
              https://github.com/rdkcentral/rdk-halif-aidl/blob/8ab4217b36a2a614384f5276343fc67226a3a236/compositeinput/current/com/rdk/hal/compositeinput/PlatformCapabilities.aidl#L60
3. dsCompositeInScaleVideo api from RDKV - is it required to be handled here or in planecontrol?
                        https://github.com/rdkcentral/rdk-halif-device_settings/blob/f91f2336d50b35c8af94e697d22ebf24d79beb77/include/dsCompositeIn.h#L234
4. Port open() close() and state transitions are not handled - is it required?
5. getStatus() - " EX_ILLEGAL_STATE if service is not ready" this line is not relevant
                       https://github.com/rdkcentral/rdk-halif-aidl/blob/8ab4217b36a2a614384f5276343fc67226a3a236/compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputPort.aidl#L85
6. getPropertyMulti() - "Entries for unsupported properties will have default-initialized values" behavior should be handled similar like other modules(return ILLEGAL_ARGUMENT even if one property not valid)
            https://github.com/rdkcentral/rdk-halif-aidl/blob/8ab4217b36a2a614384f5276343fc67226a3a236/compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputPort.aidl#L151
7. HFP - "supportedMetrics" param is not used anywhere?
              "supportedMetrics" list available in hfp, how its related to PortMetrics?
               "propertyMetadata" should be the metadata for the property, but it also has the list of metadata of "supportedMetrics"?
               https://github.com/rdkcentral/rdk-halif-aidl/blob/8ab4217b36a2a614384f5276343fc67226a3a236/compositeinput/current/hfp-compositeinput.yaml#L101</div>

Ulrond:
## Summary

Closes all remaining open gaps from #382, folded into a single generation-2 breaking change. No consumers exist yet, so all API cleanup is absorbed in one bump rather than spread across multiple.

### What's in this PR

**Gap 2 — PortProperty enum keys (Phase A + Phase B combined):**
- New `PortProperty.aidl` — canonical enum of port-scoped property keys. Named `PortProperty` (not `Property`) to match the audiomixer convention where an unprefixed `Property` is reserved for manager-level state.
- `ICompositeInputPort.getProperty(PortProperty)` / `getPropertyMulti(PortProperty[])` — enum keys, not strings.
- `ICompositeInputController.setProperty(PortProperty, PropertyValue)` / `setPropertyMulti(PropertyKVPair[])`.
- `PortCapabilities.supportedProperties` → `PortProperty[]`.
- `PlatformCapabilities.supportedProperties` → `PortProperty[]`.
- `PropertyKVPair.property` → `PortProperty` (was `String key`).
- `PropertyMetadata.key` → `PortProperty`.
- `hfp-compositeinput.yaml` — property keys are now unquoted enum identifiers.

**Gap 4 — Port lifecycle (open / close / Controller pattern):**
- New `State.aidl` — CLOSED → OPENING → READY → STARTING → STARTED → STOPPING → CLOSING (mirrors hdmiinput).
- New `ICompositeInputController.aidl` — exclusive write interface returned by `open()`: `start()`, `stop()`, `setProperty()`, `setPropertyMulti()`, `resetMetrics()`.
- New `ICompositeInputControllerListener.aidl` — controller callbacks: `onConnectionChanged()`, `onSignalStatusChanged()`, `onVideoModeChanged()`.
- `ICompositeInputPort` gains: `getState()`, `open(listener)` → `ICompositeInputController`, `close(controller)`.
- `ICompositeInputPort` loses: `setActive(boolean)` (replaced by `Controller.start()`/`stop()`), `setProperty()`, `setPropertyMulti()`, `resetMetrics()` (moved to Controller).
- `PlatformCapabilities` gains `maximumConcurrentStartedPorts` (HFP-declared, default 1).

**Listener split (aligned with hdmiinput pattern):**
- `IPortEventListener` renamed to `ICompositeInputEventListener` (module naming convention).
- Real-time A/V events (`onConnectionChanged`, `onSignalStatusChanged`, `onVideoModeChanged`) → `ICompositeInputControllerListener` (exclusive to controller owner).
- Lifecycle/observer events (`onStateChanged`, `onPropertyChanged`) → `ICompositeInputEventListener` (multi-client, registered via `registerEventListener()`).
- This matches `IHDMIInputControllerListener` (A/V events) vs `IHDMIInputEventListener` (lifecycle) exactly.

**Telemetry consolidation (resolves dual-source-of-truth):**
- Delete `PortMetrics.aidl` parcelable.
- Delete `IPortTelemetryListener.aidl`.
- `ICompositeInputPort` loses `getMetrics()`, `registerTelemetryListener()`, `unregisterTelemetryListener()`.
- `PortCapabilities` loses `metricsSupported` (inferred from presence of `METRIC_*` keys in `supportedProperties`).
- `PlatformCapabilities.FeatureFlags` loses `signalMetricsSupported` (same reason — redundant with `METRIC_*` keys).
- `PortProperty` gains `SIGNAL_QUALITY` and three new `METRIC_*` counters (`METRIC_SIGNAL_LOCK_COUNT`, `METRIC_LAST_SIGNAL_LOCK_TIME`, `METRIC_LAST_RESET_TIMESTAMP`).
- `ICompositeInputEventListener.onPropertyChanged(PortProperty, PropertyValue)` — generic property-change notification that supersedes both old telemetry callbacks.

**PortStatus cleanup:**
- Removed `signalQuality` field — signal quality is already available via `getProperty(SIGNAL_QUALITY)`. `PortStatus` is now a lean connectivity/signal/resolution snapshot; extended status and metrics go through `getProperty()`.

This aligns compositeinput with the convention already used by **videodecoder**, **videosink**, **hdmiinput** and **hdmioutput**: metrics are first-class `Property` enum keys read via the same `getProperty()` / `getPropertyMulti()` methods used for status. No separate metrics parcelable, no bespoke telemetry listener.

**Version bump:** `metadata.yaml` `0.1.0.0` → `0.2.0.0`, `generation: 1` → `2`.

### State lifecycle

```
CLOSED  ──open()──▶  OPENING ──▶  READY  ──start()──▶  STARTING ──▶  STARTED
                                    ▲                                     │
                                    └──────────stop()─── STOPPING ◀───────┘
READY   ──close()──▶ CLOSING ──▶  CLOSED
```

### Files changed

| File | Change |
|---|---|
| `PortProperty.aidl` | New (rename from `Property.aidl`, expanded with SIGNAL_QUALITY + 3 new METRIC_* keys) |
| `State.aidl` | New |
| `ICompositeInputController.aidl` | New |
| `ICompositeInputControllerListener.aidl` | New — onConnectionChanged, onSignalStatusChanged, onVideoModeChanged |
| `ICompositeInputEventListener.aidl` | New (rename from `IPortEventListener.aidl`) — onStateChanged, onPropertyChanged |
| `PortMetrics.aidl` | **Deleted** |
| `IPortTelemetryListener.aidl` | **Deleted** |
| `IPortEventListener.aidl` | **Deleted** (renamed to ICompositeInputEventListener.aidl) |
| `Property.aidl` | **Deleted** (renamed to PortProperty.aidl) |
| `ICompositeInputPort.aidl` | open/close/getState added; setActive + set* + resetMetrics + getMetrics + telemetry listener registration all removed; registerEventListener now takes ICompositeInputEventListener |
| `PortStatus.aidl` | Removed signalQuality (available via getProperty) |
| `PropertyKVPair.aidl` | String key → PortProperty |
| `PortCapabilities.aidl` | String[] → PortProperty[]; removed metricsSupported |
| `PlatformCapabilities.aidl` | String[] → PortProperty[]; added maximumConcurrentStartedPorts; removed signalMetricsSupported from FeatureFlags |
| `PropertyMetadata.aidl` | String key → PortProperty |
| `hfp-compositeinput.yaml` | Unquoted enum identifiers; supportedMetrics merged into supportedProperties; new METRIC_* keys; maximumConcurrentStartedPorts moved into platformCapabilities block |
| `CMakeLists.txt` | Add State/PortProperty/controller/listener/event-listener; remove PortMetrics/IPortTelemetryListener/IPortEventListener |
| `metadata.yaml` | 0.1.0.0 → 0.2.0.0, generation 1 → 2 |

### Template reference

Port lifecycle pattern mirrors [`IHDMIInput.aidl`](../blob/develop/hdmiinput/current/com/rdk/hal/hdmiinput/IHDMIInput.aidl) / [`IHDMIInputController.aidl`](../blob/develop/hdmiinput/current/com/rdk/hal/hdmiinput/IHDMIInputController.aidl) / [`State.aidl`](../blob/develop/hdmiinput/current/com/rdk/hal/hdmiinput/State.aidl).

Listener split mirrors [`IHDMIInputControllerListener.aidl`](../blob/develop/hdmiinput/current/com/rdk/hal/hdmiinput/IHDMIInputControllerListener.aidl) (A/V events to controller owner) vs [`IHDMIInputEventListener.aidl`](../blob/develop/hdmiinput/current/com/rdk/hal/hdmiinput/IHDMIInputEventListener.aidl) (lifecycle to any observer).

Telemetry-in-Property-enum pattern mirrors [`videosink/Property.aidl`](../blob/develop/videosink/current/com/rdk/hal/videosink/Property.aidl) (7 METRIC_* keys), [`videodecoder/Property.aidl`](../blob/develop/videodecoder/current/com/rdk/hal/videodecoder/Property.aidl), [`hdmiinput/Property.aidl`](../blob/develop/hdmiinput/current/com/rdk/hal/hdmiinput/Property.aidl), [`hdmioutput/Property.aidl`](../blob/develop/hdmioutput/current/com/rdk/hal/hdmioutput/Property.aidl).

## Test plan

- [ ] All CI checks pass (Signature, blackduck, copyright)
- [ ] `PortProperty` enum values match all keys declared in `hfp-compositeinput.yaml`
- [ ] `ICompositeInputPort.open()` / `close()` / `getState()` / `getStatus()` present and correctly typed
- [ ] `ICompositeInputController.start()` / `stop()` / `setProperty()` / `setPropertyMulti()` / `resetMetrics()` present
- [ ] `ICompositeInputControllerListener`: `onConnectionChanged()`, `onSignalStatusChanged()`, `onVideoModeChanged()` present
- [ ] `ICompositeInputEventListener`: `onStateChanged()`, `onPropertyChanged()` present
- [ ] `PortMetrics.aidl`, `IPortTelemetryListener.aidl`, `IPortEventListener.aidl` removed
- [ ] `PortCapabilities.metricsSupported` removed; `supportedProperties` is `PortProperty[]`
- [ ] `PlatformCapabilities.maximumConcurrentStartedPorts` present; `signalMetricsSupported` removed
- [ ] `PortStatus` has no `signalQuality` field
- [ ] `hfp-compositeinput.yaml`: `maximumConcurrentStartedPorts` inside `platformCapabilities` block
- [ ] `metadata.yaml` version `0.2.0.0`, generation `2`
- [ ] Root CMake build parses all new AIDL files without errors


compositeinput/current/com/rdk/hal/compositeinput/IPortEventListener.aidl
```diff
@@ -19,69 +19,76 @@
...
+ * Multiple listeners can be registered on the same port via
+ * ICompositeInputPort.registerEventListener(). Observers do not need to own
+ * the controller — this is the multi-client observation channel for UI,
+ * diagnostics, telemetry, and other watchers.
  */
 @VintfStability
 oneway interface IPortEventListener
 {
     /**
-     * Callback for port connection status changes.
```
thanushreevani, line 39: All these callback API's has to be in Controller listener, similar like other modules.

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputControllerListener.aidl
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Called when the port state changes.
+     *
+     * Fired for all transitions driven by open(), close(), start(), and stop().
+     *
+     * @param[in] oldState  The state being left.
+     * @param[in] newState  The state being entered.
+     */
+    void onStateChanged(in State oldState, in State newState);
```
thanushreevani: This callback API's has to be in Event listener, similar like other modules.

compositeinput/current/com/rdk/hal/compositeinput/State.aidl
```diff
@@ -0,0 +1,69 @@
...
+ */
+@VintfStability
+@Backing(type="int")
+enum State {
+
+    /**
+     * Initial state. The port is not open; no controller exists.
+     * Entered on service connection or after a successful close().
+     */
+    CLOSED = 0,
```
thanushreevani, line 36: UNKNOWN state tobe added similar like other module

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputPort.aidl
```diff
@@ -19,206 +19,178 @@
...
-     * @exception binder::Status EX_ILLEGAL_ARGUMENT if propertyKey is null or empty.
+     * Returns a polled snapshot of connection state, signal status, and detected
+     * video mode. Use in conjunction with IPortEventListener for async updates.
      *
-     * @see PropertyValue, getCapabilities(), setProperty(), getPropertyMulti()
+     * @returns PortStatus parcelable with current port state.
+     * @see PortStatus, IPortEventListener
      */
-    @nullable PropertyValue getProperty(in @utf8InCpp String propertyKey);
+    PortStatus getStatus();
```
thanushreevani, line 116: This API to be moved to Controller interface ?

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputController.aidl
```diff
@@ -0,0 +1,120 @@
...
+     * Resets telemetry metric counters for this port.
+     *
+     * Clears all accumulated METRIC_* values on this port back to their initial
+     * state, and updates METRIC_LAST_RESET_TIMESTAMP to the current wall-clock time.
+     *
+     * @exception binder::Status EX_ILLEGAL_STATE if port is not in STARTED state.
+     *
+     * @see PortProperty, ICompositeInputPort.getProperty()
+     */
+    void resetMetrics();
```
thanushreevani, line 120: Is this API required as the Metric values will be updated/retrieved using get/set property

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputControllerListener.aidl
Ulrond: this is the controllers event listener, which we have used in other modules.
the question is other than controllers is anyone else interested in the state, ( that's what the general event listener is for ).

compositeinput/current/com/rdk/hal/compositeinput/State.aidl
Ulrond: that's incorrect, each module has their own states, UNKNOWN or any state should only exist if the module things' based on the design if it's required.

this module shouldn't have UNKNOWN state there is always a KNOWN state.

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputPort.aidl
Ulrond: controllers are for Set State not GetSTatus anywhere currently where we've got getState in a controller then potentially that's a problem.

compositeinput/current/com/rdk/hal/compositeinput/IPortEventListener.aidl
Ulrond: not sure on what your saying..

we've got two types of listener types today.

Ones for the controllers, where only the controller get's calllbacks, that's the controllerListener()
And a general eventListenerI() anyone can register for.

If any module who is not a controller wants to listen that's allowed.

Ulrond: You're right — revisited against the hdmiinput pattern and the split in this PR was inverted. Fixed in commit 1218d3e6:

- `onConnectionChanged`, `onVideoModeChanged` → moved to `ICompositeInputControllerListener` (matches `IHDMIInputControllerListener.onConnectionStateChanged` / `onVIChanged`)
- `onSignalStatusChanged` → kept on controller listener only (was duplicated on both — removed from event listener)
- `onStateChanged` → moved **to** event listener (matches `IHDMIInputEventListener.onStateChanged`)
- File renamed `IPortEventListener` → `ICompositeInputEventListener` to match module naming

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputControllerListener.aidl
Ulrond: Yes, agreed. `onStateChanged` belongs on the event listener so non-owners can observe port lifecycle without exclusive ownership (same reasoning as `IHDMIInputEventListener.onStateChanged`). Moved in commit 1218d3e6.

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputController.aidl
Ulrond: Thanushreevani: the `resetMetrics()` API on `ICompositeInputController` is still needed distinctly from `setProperty()`. Metric values themselves are read/updated via `getProperty` (they're `METRIC_*` PortProperty keys), but there's no per-key setter to clear a counter — and the reset is atomic across all metrics plus updates `METRIC_LAST_RESET_TIMESTAMP` in one operation. Splitting that into a series of `setProperty` calls would lose atomicity and require exposing the metric keys as writable, which contradicts their readOnly metadata. Will add this clarification to the `resetMetrics()` doc.

compositeinput/current/com/rdk/hal/compositeinput/ICompositeInputControllerListener.aidl
```diff
@@ -0,0 +1,84 @@
...
+oneway interface ICompositeInputControllerListener {
+
+    /**
+     * Called when the port connection state changes.
+     *
+     * Fired by hardware detection of cable presence. Indicates physical
+     * connection only, not signal validity.
+     *
+     * Always fires at least once during the OPENING transition to report
+     * the current connected/disconnected state before READY is reached.
```
thanushreevani, line 50: Is below line applicable for compositeIn onConnectionChange event also
     * - **Only on actual HPD changes** during `STARTING`, `STARTED`, and `STOPPING` states
     *   (i.e., when cable is physically connected/disconnected or compositeInput device is powered on/off).
Note: referred from HdmiOutput

Ulrond: Good question. For **input** modules the callback fires in **two** cases:

1. **Always during OPENING** — to report the initial connected/disconnected state before READY is reached. This matches `IHDMIInput.open()` which states: *"The `IHDMIInputControllerListener.onConnectionStateChanged()` callback is always fired during the `OPENING` state to indicate the current connected/disconnected state."*

2. **On actual physical cable changes** during `READY`, `STARTING`, `STARTED`, `STOPPING` — debounced by the HAL, same as HDMI input.

The HDMI **output** pattern you referenced (`onConnectionChange` only on actual HPD changes, no guaranteed initial report) is different because output modules don't need to discover whether a sink is connected during open — the STB drives the output regardless. Input modules need that initial snapshot so the client knows whether to proceed to `start()`.

So yes, both lines apply to compositeInput:
- Guaranteed during OPENING (initial report)
- On physical changes thereafter (hot plug/unplug)

Dhivyailangovan (commented):
Hi,
1. I noticed that PortStatus.aidl is available for the composite input, but it is not present for HDMI. Should we also add this support to HDMI?Or is it intentional that PortStatus.aidl is only in composite? If not, should we instead remove it from composite?
2. There are some naming differences between HDMI and Composite—
for example, HDMI uses "locked state" while Composite uses "stable state" for signal state. File names also differ between the two. Do we need to align these naming conventions between HDMI and Composite? or is this intentional?

Ulrond:
@Dhivyailangovan good questions, both are intentional:

**1. PortStatus.aidl** — composite has it, HDMI doesn't. This is a generation-2 design improvement. `PortStatus` bundles `connected`, `signalStatus`, and `videoResolution` into a single polled snapshot returned by `getStatus()` — one IPC call instead of three. HDMI input exposes these individually via `IHDMIInputController.getConnectionState()`, `getSignalState()`, etc. Both approaches work. Since composite is new, it uses the cleaner pattern. Retroactively adding `PortStatus` to HDMI would be a separate enhancement if desired, but isn't required for consistency — the per-field getters on HDMI are functionally equivalent.

**2. Naming: LOCKED vs STABLE** — different signal domains, different terminology. Composite analog signals have `STABLE` (analog sync lock is a stability measurement); HDMI digital has `LOCKED` (TMDS link lock is a binary state). Renaming to match would be technically incorrect — they describe different physical phenomena. File naming differences (`IHDMIInput` vs `ICompositeInputPort`) follow the same module-specific naming pattern used across the repo.

Ulrond:
@Dhivyailangovan follow-up on the PortStatus question — after looking at this more closely, there was a concrete overlap issue worth fixing.

**Decision: keep PortStatus on composite, don't add to HDMI, fix the data overlap.**

`PortStatus.signalQuality` was duplicating `PortProperty.SIGNAL_QUALITY` — two read paths to the same data, meaning the HAL has to keep them in sync and callers pick arbitrarily between them. Removed `signalQuality` from `PortStatus` in commit 1487fcdf.

**Clean split is now:**
- `getStatus()` → lean snapshot: connected, active, signalStatus, detectedResolution (one IPC call for the "is it working?" question)
- `getProperty()` → everything else: signal quality, signal strength, telemetry metrics (extensible, no duplication)

**Why not add PortStatus to HDMI:** HDMI's status surface is much wider — VIC, VRR, HDCP version/status, InfoFrames, EDID. A flat parcelable would either be incomplete or bloated. HDMI already has the Property system for extensible status reads, plus individual getters on the controller. Adding a partial snapshot creates the same overlap problem we just fixed on composite, but worse.

srinivasgtl (dismissed):
approved.

PR #454, opened 2026-04-28
sensor: Fix build issues by adding CMakeLists and VINTF stability

jfreeda:
sensor: Add missing CMakeLists.txt and @VintfStability annotations
Add sensor/current/CMakeLists.txt, which was missing and caused CMake to abort with "sensor does not have version `current`" when building the vcomponent. The file lists all motion and thermal AIDL sources and invokes compile_aidl().

Add missing @VintfStability annotations to IThermalSensor.aidl and IThermalEventListener.aidl. The AIDL compiler requires this annotation for interfaces compiled with --stability=vintf.

### DIFF
--- a/.github/workflows/update-rag-report.yml
+++ b/.github/workflows/update-rag-report.yml
@@ -19,11 +19,21 @@ jobs:
         run: bash scripts/generate_rag_report.sh
 
       - name: Create Pull Request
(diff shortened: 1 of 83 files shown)


jfreeda:
Creating another PR with base of 
feature/427-motion-sensor-controller-pattern

Issue #474, opened 2026-05-01
Task: [motion_sensor] Address post-merge Copilot review on PR #430


Ulrond:
PR #430 (motion sensor controller pattern) is merged. Copilot raised 7 follow-up comments on the merged code. This issue tracks fixing them.

## Real bugs

- **`motion_sensor.md` Initialization section** — still says `serviceName = "MotionSensorManager"` while the requirements table at `HAL.MOTION.10` already uses `"sensor.motion"`. Pick one (the AIDL constant is `"sensor.motion"`).

## Design questions raised by Copilot (need rationale, may not require code change)

- **`IMotionSensorController.start()/stop()` return boolean** — Copilot noted this differs from other modules' controllers (e.g. `hdmiinput` uses `void start()` + binder exceptions). The boolean signals hardware-init failure separately from EX_ILLEGAL_STATE — already documented in the controller header. Worth a comment reply explaining the dual-signal pattern was intentional and confirmed in the original review thread with jfreeda.

- **`LastEventInfo.timestampNs` naming** — Copilot suggested the name doesn't communicate the time base (CLOCK_MONOTONIC). Options: rename to `monotonicTimestampNs` (breaking) or add a stronger doc note. Pre-freeze, so renaming is cheap if we want clarity.

## Already addressed (Copilot looking at stale view)

- Interface Definition link text — already fixed to point at `sensor/current/com/rdk/hal/sensor/motion`
- `HAL.MOTION.10` service name — already correct in the table
- `motion_sensor.md` sequence diagram `onStateChanged(STOPPED, STOPPED)` — already replaced with a Note

## Refs

- PR #430 (merged)
- Copilot comments on PR #430 dated 2026-04-30

Ulrond:
Closing — over-tracked. The only actionable item is a one-line doc fix on PR #475. Design-question replies on PR #430 don't need a tracking issue.

PR #430, opened 2026-04-22
feat(motion_sensor): adopt controller pattern with StartConfig and diagnostics (#427)

Ulrond:
## Summary

Adopts the controller pattern for the motion sensor HAL, aligning with hdmiinput, compositeinput, and other stateful modules. Resolves #427.

- **`IMotionSensorController`** — exclusive write controller returned by `open()`: `start(StartConfig)`, `stop()`, `getStartConfig()`, `getLastEventInfo()`, sensitivity, deep sleep autonomy, active windows
- **`IMotionSensorControllerListener`** — `onStateChanged(oldState, newState)`, `onActiveWindowEntered()`, `onActiveWindowExited()`
- **`StartConfig`** parcelable — bundles all `start()` parameters with a getter for config retrieval
- **`LastEventInfo`** parcelable — diagnostic: last event mode + timestamp
- **`IMotionSensor`** refactored — gains `open(listener)`/`close(controller)`, keeps `getCapabilities()`/`getState()`/event listener registration, all mutators moved to controller
- Fixed missing `@VintfStability` on `IMotionSensor`, `IMotionSensorManager`, `IMotionSensorEventListener` (pre-existing omissions)
- HFP timing section renamed to `defaultStartConfig` matching parcelable field names

### Breaking change

`start()` moves from `IMotionSensor` (4 params) to `IMotionSensorController.start(StartConfig)`. All configuration/mutation methods move to the controller. Pre-freeze, no consumers — no migration shim needed.

## Test plan

- [ ] AIDL compiles cleanly (verified locally)
- [ ] `IMotionSensor.open()` / `close()` present and correctly typed
- [ ] `IMotionSensorController.start(StartConfig)` / `stop()` / `getStartConfig()` / `getLastEventInfo()` present
- [ ] `IMotionSensorControllerListener.onStateChanged()` / `onActiveWindowEntered()` / `onActiveWindowExited()` present
- [ ] `IMotionSensorEventListener.onEvent()` unchanged
- [ ] `IMotionSensor` no longer has start/stop/setSensitivity/setActiveWindows etc.
- [ ] All interfaces have `@VintfStability`
- [ ] CMakeLists.txt includes all new files
- [ ] HFP `defaultStartConfig` fields match `StartConfig.aidl` field names

PR #475, opened 2026-05-01
docs(motion_sensor): fix stale serviceName reference

Ulrond:
## Summary

motion_sensor.md Initialization section still said \`serviceName = \"MotionSensorManager\"\` while the requirements table (HAL.MOTION.10) already uses the correct \`\"sensor.motion\"\` matching the AIDL constant. This PR brings the two references into agreement.

Found in Copilot post-merge review on PR #430. Tracked in #474.

The other items in #474 are design-question replies (start/stop boolean rationale, LastEventInfo naming) that can be addressed separately or as comments on the original PR.

## Test plan

- [ ] Doc text matches \`IMotionSensorManager.serviceName = \"sensor.motion\"\`

### DIFF
--- a/docs/halif/sensor/current/motion/motion_sensor.md
+++ b/docs/halif/sensor/current/motion/motion_sensor.md
@@ -93,7 +93,7 @@ Each sensor type has different characteristics (range, field-of-view, sensitivit
 ## Initialization
 
 Vendors shall provide a `hal-sensor-motion.service` systemd unit to launch the Motion HAL.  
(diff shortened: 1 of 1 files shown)


Ulrond:
Closing — Copilot was wrong. The Initialization section refers to the registered service name ("MotionSensorManager"), which is distinct from the AIDL string constant `IMotionSensorManager.serviceName = "sensor.motion"`. The original doc was correct.

Issue #603, opened 2026-06-10
Task: Transform Westeros onto IGraphicsFbProvider (retire per-vendor GL backends)


Ulrond:
## Goal
Produce an architecture/transformation document describing how Westeros (and its successor compositor) is upgraded to consume the merged `IGraphicsFbProvider` interface on `IPlaneControl`, replacing the per-vendor Westeros GL backends (`westeros-gl-brcm`, `westeros-gl-drm`, `westeros-soc-mtk`, …) with a single platform-agnostic implementation.

The HAL contract already exists and merged (`planecontrol/current`): `getNativeDisplayHandle()` + `getEGLPlatformType()` (EGL setup), `createGraphicsFb()`/`destroyGraphicsFb()` (DMA-buf alloc), `commitGraphicsFb()` + `IGraphicsFbProviderListener.onGraphicsFbReleased()` (present/recycle). The compositor needs only: an EGL display+context, renderable DMA-buf buffers, and a commit/release lifecycle — all three are provided.

## Scope
- New `docs/transformation/` area; first doc `westeros_graphics_fb_transform.md`.
- Corrects an earlier draft (`graphics-frame-provider-design-rationale.md`, from the #402/PR #423 work) whose API names predate the merged `*Fb*` naming, and which placed `format`/`modifier` per-frame rather than in `GraphicsFbCapabilities` (where they are opaque, EGL-known values).

## Corrections the doc must apply vs current AIDL
- `IGraphicsFrameProvider`→`IGraphicsFbProvider`; `GraphicsFrameInfo`→`GraphicsFbInfo`; `GraphicsFrameCapabilities`→`GraphicsFbCapabilities`; `createGraphicsFrame`→`createGraphicsFb` (and destroy/commit); `frameId`→`graphicsFbId`; `onFrameReleased`→`onGraphicsFbReleased(oldGraphicsFbId, elapsedRealtimeNanos)`.
- `getGraphicsFbProvider(planeResourceIndex, listener)` is `@nullable`.
- `format`/`modifier` come from `getCapabilities()` and are **opaque** (not mandated DRM fourcc); read them once, not per-frame.
- The returned fd may be shared across buffers — import keyed on `offset`.

## Related
Relates to #329 (PlaneControl controller audit), #33 (Graphics Allocation / Gralloc review), #62 (VSI Graphics specification).

PR #621, opened 2026-06-17
feat(planecontrol)!: drop EGL-display methods from IGraphicsFbProvider + transform guide (#603, #620)

Ulrond:
## Summary

A single branch carrying the interface change and the transformation guide.

**Interface (`planecontrol/current/IGraphicsFbProvider.aidl`):** remove `getNativeDisplayHandle()` and `getEGLPlatformType()`. A native display is a process-local, vendor-specific object and cannot cross the Binder boundary. Because presentation is via `commitGraphicsFb()` (offscreen FBO render, not a window surface), the compositor needs no native display — EGL display/context setup stays client-side in the platform EGL layer. The interface carries only buffer allocation and presentation.

**Doc (`docs/transformation/westeros_graphics_fb_transform.md`):**
- Scope the HAL contract to **buffer allocation + presentation**.
- Rewrite the EGL-init narrative for offscreen, client-side setup.
- Keep it **vendor-neutral**: describe the universal DMA-buf mechanism (open DRM/GBM example, generic "platforms without GBM"), and reference only the public Westeros backends — no proprietary vendor API or firmware detail.

Resolves the docs + interface portion of #603.

## Notes for reviewers

- `current/` tracks AIDL source only (bindings are generated at build/freeze), so there are no generated files to regenerate here.
- This is a **breaking** removal on `current`. `IGraphicsFbProvider` is present in the `0.1.0.0` frozen snapshot but not `0.2.0.0`/`0.3.0.0`; freeze/compat handling is left to the release process.
- Background discussion: #620.

### DIFF
--- a/docs/transformation/westeros_graphics_fb_transform.md
+++ b/docs/transformation/westeros_graphics_fb_transform.md
@@ -1,0 +1,458 @@
+# Transforming Westeros onto IGraphicsFbProvider
+
+## Purpose
+
+This document describes how the Westeros compositor (and its successor) is
+transformed to render through the Plane Control HAL's graphics frame buffer
+interface, `IGraphicsFbProvider`, replacing the per-vendor Westeros GL
+backends with a single, platform-agnostic implementation.
+
+The HAL contract is vendor-agnostic and complete: it covers EGL display
+setup, DMA-buf buffer allocation, and presentation. The compositor needs
+exactly those three things, so all SoC-specific graphics code moves below
+the HAL boundary and is implemented once per platform.
+
+Related interface: [`planecontrol/current`](../../planecontrol/current/docs/plane_control.md).
+Related issues: #603 (this work), #329 (PlaneControl controller audit),
+issue #33 (Graphics allocation / Gralloc review), #62 (VSI Graphics specification).
+
+---
+
+## 1. The interface the compositor consumes
+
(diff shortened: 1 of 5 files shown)


Ulrond:
Thanks @Copilot — responses:

1. **EGL attribute-list type (`docs:175`)** — fixed in 1a4c3a74: the example now uses `EGLint attrs[]` for `eglCreateImageKHR`.

2. **`getNativeDisplayHandle()` → `void*` cast note (`docs:159`)** — no longer applicable. Both `getNativeDisplayHandle()` and `getEGLPlatformType()` have been removed from the interface, and the snippet now creates an offscreen `EGLDisplay` client-side (no handle crosses Binder). This is the resolution of the cross-process concern you flagged — see #620.

3. **Breaking change vs `0.1.0.0` (`IGraphicsFbProvider.aidl:108`)** — agreed this is an ABI/API break against the released snapshot and needs to be handled as deliberate interface evolution, not an in-place delete. Flagging for a maintainer decision on the mechanism (deprecate-in-place vs new interface version). Tracking separately rather than silently landing it here.

Separately, the `onGraphicsFbReleased` timestamp parameter (renamed `elapsedRealtimeNanos` → `timestampNs` here) has open questions — necessity, and server-side population (the POC never populated it) — now tracked in **#622**.

shafi12 (approved):

Issue #624, opened 2026-06-22
Bug: build_binder.sh: fails in OE ARM cross-compile environment due to hardcoded BUILD_HOST_AIDL=ON and missing OECORE env suppression


Anbukannadhasan:
### Problem/Opportunity

build_binder.sh has two compounding failures when invoked from an OpenEmbedded (OE) ARM cross-compile environment:

**Problem 1** - BUILD_HOST_AIDL=ON is hardcoded regardless of active compiler Lines 250 and 257 of build_binder.sh unconditionally pass -DBUILD_HOST_AIDL=ON to cmake. When CC/CXX are set to an ARM cross-compiler (as OE does), cmake compiles the AIDL host tool binary using the ARM cross-compiler. This fails in two ways:
- The resulting binary is ARM - it cannot execute on the x86 build host. cmake fails immediately when it tries to run the AIDL tool to verify it.
- Even when trying with an x86 compiler, the OE native sysroot prepends its own m4 to PATH. That older m4 rejects the --gnu flag that bison passes when building the AIDL parser grammar, producing a cryptic m4 error mid-build that has nothing to do with user code.

**Problem 2** - OECORE_* env vars are not suppressed before cmake configure linux_binder_idl/CMakeLists.txt line 174 auto-enables BUILD_ENV_YOCTO=ON whenever OECORE_NATIVE_SYSROOT or OECORE_TARGET_SYSROOT are present in the environment. When BUILD_ENV_YOCTO=ON, all cmake install() rules are disabled, so no libraries or headers are ever installed to out/target/. build_binder.sh does not unset these variables before running cmake configure, so calling it inside an active OE shell results in a build that appears to succeed but produces no output.

### Steps to reproduce

  1. Source an OE ARM toolchain environment (sets CC, CXX, CFLAGS, OECORE_NATIVE_SYSROOT, OECORE_TARGET_SYSROOT)
_sc docker run rdk-kirkstone /bin/bash
source /opt/toolchains/rdk-glibc-x86_64-arm-toolchain/environment-setup-armv7vet2hf-neon-oe-linux-gnueabi_
  2. cd rdk-halif-aidl
  3. ./build_binder.sh

### Expected Behavior

  - When CC points to an ARM cross-compiler, build_binder.sh detects this and passes -DBUILD_HOST_AIDL=OFF so only the binder runtime libraries are built (skipping the AIDL host tool, which requires an x86 compiler)
[...]

PR #628, opened 2026-06-22
fix(build): guard dev wrapper scripts against cross/OE env + document 3rd-party integration

Ulrond:
Refs #624

## Background

#624 reports `build_binder.sh` failing in an OpenEmbedded ARM cross environment two ways: the host AIDL tool gets cross-compiled (can't run on the build host), and `OECORE_*_SYSROOT` makes the binder CMake set `BUILD_ENV_YOCTO=ON`, which disables every `install()` rule so `out/target` ends up empty.

**Both behaviours are correct for a production build invoked directly through CMake.** The actual problem is that the `build_*.sh` wrapper scripts — which are documented as developer/architecture-only tools ([BUILD.md](build-tools/linux_binder_idl/BUILD.md): *"Yocto/BitBake recipes must NOT use these wrapper scripts"*) — were run in an environment they were never meant for. The HAL libraries build from committed generated C++, so the AIDL **host** tool is never needed on a target at all.

So this is not a code adaptation bug — the fix is to **enforce and document the boundary**.

## Changes

- **`dev_env_guard.sh`** (new) — shared guard sourced by `build_binder.sh`, `build_modules.sh`, `build_interfaces.sh`. If it detects a cross `CC` (target triple ≠ host) or `OECORE_*_SYSROOT`, it aborts with a clear message pointing to the direct-CMake recipe. `clean`/help paths bypass it so they stay usable anywhere; `BINDER_ALLOW_CROSS_ENV=1` overrides.
- **`docs/standards/build_integration.md`** (new, wired into the nav) — the integration contract for third-party/Yocto build systems: two-stage build via direct CMake, required variables, BitBake recipe pattern, and why the wrapper scripts are not used.
- **README** — production-build section now states the contract and links the new doc.

## Testing

- `bash -n` on all four scripts — clean.
- Guard unit checks: native (no `CC`/`OECORE`) → allowed; `OECORE_*` set → blocked; cross `CC` → blocked; `BINDER_ALLOW_CROSS_ENV=1` → allowed.
- Wrapper smoke tests: `build_modules.sh --help` bypasses the guard (exit 0); `build_modules.sh <module>` and `build_interfaces.sh sdk` in a simulated OE env abort before doing any work with the pointer message.

---
### Update: reference recipes added
Folded in copy-me integration recipes so the guide ships **instructions + examples**:
- `docs/standards/examples/rdk-halif-aidl.bb` — BitBake reference recipe.
- `docs/standards/examples/rdk-halif-aidl.yaml` — Bob Build Tool reference recipe.

Both are thin wrappers over the same direct-CMake invocation, demonstrating the build is build-system-agnostic. Linked from `build_integration.md`; marked as reference templates (pin to a release tag, adapt toolchain/sysroot), not CI-verified — I have no Yocto/Bob environment to execute-test them.

### DIFF
--- a/docs/standards/examples/rdk-halif-aidl.bb
+++ b/docs/standards/examples/rdk-halif-aidl.bb
@@ -1,0 +1,57 @@
+#/**
+# * Copyright 2026 RDK Management
+# *
+# * Licensed under the Apache License, Version 2.0 (the "License");
+# * you may not use this file except in compliance with the License.
+# * You may obtain a copy of the License at
+# *
+# * http://www.apache.org/licenses/LICENSE-2.0
+# *
+# * Unless required by applicable law or agreed to in writing, software
+# * distributed under the License is distributed on an "AS IS" BASIS,
+# * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# * See the License for the specific language governing permissions and
+# * limitations under the License.
+# *
+# * SPDX-License-Identifier: Apache-2.0
+# */
+#
+# REFERENCE BitBake recipe for the RDK HAL AIDL interface libraries.
+#
+# This is a copy-me template, NOT a recipe consumed by this repository. Drop it
+# into your meta-layer, pin SRCREV/branch to a released tag, and adjust as
+# needed. It builds the committed, pre-generated C++ via direct CMake — no AIDL
+# compiler or Python is required on the build host (see
+# docs/standards/build_integration.md for the full contract).
+
+SUMMARY = "RDK HAL AIDL interface libraries (Polaris HAL)"
+HOMEPAGE = "https://github.com/rdkcentral/rdk-halif-aidl"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=86d3f3a95c324c9479bd8986968f4327"
+
+SRC_URI = "git://github.com/rdkcentral/rdk-halif-aidl.git;protocol=https;branch=develop"
+# Pin to a released tag for reproducible builds, e.g.:
+#   SRCREV = "0.21.0"
+SRCREV = "${AUTOREV}"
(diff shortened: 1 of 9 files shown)


Ulrond:
Addressed Copilot review feedback (pushed):
- **Guard `$0` when sourced** → now uses `BASH_SOURCE[1]`, so the message names the actual wrapper.
- **Guard doc pointer / BUILD.md links** → repointed from `build-tools/linux_binder_idl/BUILD.md` (a gitignored clone, absent from a fresh checkout) to the in-tree `docs/standards/build_integration.md` and the upstream linux_binder_idl BUILD guide.
- **`versions_released.yaml` wording** → clarified it's consumed by `build_modules.sh`, not the top-level CMake; a cohort build iterates versions itself.
- **BitBake `do_install` path** → replaced the incorrect `${B}/out/target` copy with the `cmake` class (EXTRA_OECMAKE + class-managed install).
- **`build_binder.sh clean` not bypassing the guard** → intentional and now documented: unlike the sibling scripts' pure clean, `clean` here means *force rebuild*, so it's a build op that correctly needs a host toolchain.

docs/standards/examples/rdk-halif-aidl.bb
```diff
@@ -0,0 +1,57 @@
+#/**
```
mhughesacn: Hi @Ulrond : Please remove the Apache header here as by convention recipes in RDK meta layers (and elsewhere) usually don't have a header and are licensed as MIT for the most part.

Ulrond: Thanks @mhughesacn — removed the Apache header from both reference recipes (the .bb and the Bob .yaml); they're licensed via the repo per the meta-layer convention. Done in 4f96d071.

Ulrond:
Copilot review: fixed the SRCREV pinning example (SHA, not tag name). The `build_interfaces.sh` debug-flags example thread is outdated (line since changed) — resolved.

Issue #633, opened 2026-06-23
feat(versioning): trustworthy frozen interface versioning — real + honest getInterfaceVersion()/getInterfaceHash()


Ulrond:
## Goal

Frozen snapshots must expose a **real and honest** interface identity: `getInterfaceVersion()` / `getInterfaceHash()` return the actual version + contract hash (not `1` / `notfrozen`), **and** those numbers are guaranteed to reflect true ABI compatibility.

Consolidates the former **#658** (change-class enforcement gate) — the two are halves of one capability: *produce* the version/hash, and *guarantee* it is honest. Without the gate, the `server >= client` compat check is advisory, not guaranteed.

## Design (reconciled to linux_binder_idl#32)

Earlier drafts of this ticket and PR #634 proposed different encodings (positional int `0.2.0.1→20001`; index-in-released-list). Those are superseded by the **ABI-only** scheme decided in linux_binder_idl#32:

- **`getInterfaceVersion()` = ABI-only ordinal** — encodes `generation`.`minor` only; the **patch/doc field is dropped**, so a doc-only bump does not change it. Monotonic; higher = additive-superset.
- **`getInterfaceHash()` = normalized-API hash** — hash of the canonical API dump (methods/signatures/fields/enums), comment/doc/format-insensitive; changes iff the ABI changes.
- **Compatibility predicate** (consumer): `generation(server) == generation(client) && server >= client` — well-defined under the ABI-only encoding.

## Division of labour (who does what)

1. **Generator — linux_binder_idl#32** (builds on #33, merged 2.4.0): *emits* both signals from the AIDL. This is where the version/hash are computed. **Not** `release.sh`.
2. **`release.sh` — this ticket (PR #634, re-scoped):** only *sets the 4-field release version* in the snapshot's `interface.yaml`; the generator emits from it. Also: retrofit the missing/empty `.hash` for the ~10 existing cohort snapshots. **Drop** `_contract_hash` and the positional `_snapshot_version_int` (both belong in the generator).
3. **Change-class enforcement gate (was #658), at `release.sh create_snapshot`:** for each component,
   - canonical surface dump of the **previous released** vs the **about-to-be-frozen** AIDL,
   - classify the structural diff as **additive** / **breaking** / **none**,
   - assert the planned bump matches: breaking ⇒ generation increments; additive ⇒ minor/patch, generation unchanged; none ⇒ no snapshot,
   - **fail the release** with a clear diff on mismatch.
[...]

Issue #33, opened 2025-02-12
Task: Review Graphics Allocation (Gralloc)


Ulrond:
### Goal

Need to review Graphics allocation and check if we can add that as an AIDL interface and what is required for it. 

 
[Gralloc](https://android.googlesource.com/platform/hardware/interfaces/+/ca31d800199bd0624bc5196bb69d5d4c97e2f8e9/graphics/allocator/aidl/aidl_api/android.hardware.graphics.allocator/2/android/hardware/graphics/allocator/BufferDescriptorInfo.aidl)

Ulrond:
Linked to the new allocator direction: #668 (conceptual data-source HAL) / PR #669 (datasource module — BufferDescriptor + IDataSource).

Issue #32, opened 2025-02-12
Feature: New Future / Interfaces


Ulrond:
### Problem/Opportunity

- Scope out new interfaces for review

### Proposed Solution

- Review new interfaces

### Acceptance Criteria (Optional)

- List out tickets for any new interfaces that need to be defined

Ulrond:
Umbrella closed — the anticipated interfaces are now scoped RAG components (broadcast, r4ce, ffv, firmwareupdate); child tickets remain open.

PR #634, opened 2026-06-23
feat(release): bake real interface version + contract hash into snapshots at freeze

Ulrond:
Refs #633 — consumer-side half of linux_binder_idl #32 / PR #33.

## Problem
`getInterfaceVersion()` / `getInterfaceHash()` return `1` / `notfrozen` for every snapshot, so a client can't tell which interface version a service exposes. Snapshots are **compile-only** (their CMakeLists glob `src/*.cpp`), so this can only be fixed at **freeze time**.

## Change (`create_snapshot`)
Before generating, it now:
- derives the **interface VERSION from the X.Y.Z.W digits** — the int32 is the version with dots removed (base-10): `0.2.0.0 → 0200 → 200`, `0.3.0.0 → 300`, `0.1.0.1 → 101`;
- computes the **contract hash** (`sha1` of sorted per-file `.aidl` sha1sums + the `version_for_hashgen` label), matching the toolchain's integrity check;
- **stamps `current/`** with `version: <int>` + `.hash`, generates (the version-aware generator emits them), copies into the snapshot, **restores `current/`**, and records the snapshot's `.hash`.

## Proof (tested end-to-end, audiodecoder 0.2.0.0)
Snapshot emitted `static const int32_t VERSION = 200;` and `HASH = "bd961e2c…"`, imports still resolved (generator decouples emit-version from resolution-version), library built. Encoding spot-check: `0.1.0.0→100, 0.2.0.0→200, 0.3.0.0→300, 0.1.0.1→101`.

## Notes
- int32 can't carry a leading zero, so `0.2.0.0` reads back as **200** (zero-pad to 4 digits to display `0200`). Assumes single-digit components (true for all current versions); a wider fixed-field scheme would be needed if a component reaches 10.
- With today's generator the `.hash` already flows → HASH correct; the VERSION digit-encoding activates once **linux_binder_idl [#33](https://github.com/rdkcentral/linux_binder_idl/pull/33)** is merged and `binder_sdk.version` is bumped.
- Follow-on (separate): retrofit the existing cohort snapshots (10 lack a `.hash`) by re-freezing.


Ulrond:
Copilot review: the three flagged items (SHA-1→SHA-256 contract hash, strict 4-field version validation, and the positional version-int naming) were already addressed in follow-up commits — resolving the now-outdated threads.

Ulrond:
## Re-scope (decided 2026-07-04)

The **ABI signals belong in the generator**, not here — see [linux_binder_idl#32](https://github.com/rdkcentral/linux_binder_idl/issues/32) (now on the 2.5.0 milestone), which owns:

- `getInterfaceVersion()` = **ABI-only** ordinal (`generation`.`minor`; the patch/doc field is dropped, so a doc-only bump doesn't move it).
- `getInterfaceHash()` = **normalized-API** hash (contract-level; comment/doc/format-insensitive), computed by the tool that understands the AIDL.

That makes two changes here:

1. **Drop `_contract_hash`.** `release.sh` should not define a hash. A byte-level sha256 of the `.aidl` files (a) is a *second* hash definition parallel to the generator's, and (b) changes on pure reformatting/comments — a false "contract changed" signal. The real hash comes from the generator (#32).
2. **Drop `_snapshot_version_int`'s 4-field positional encoding.** It encodes all four fields (`0.2.0.1 → 20001`), so `getInterfaceVersion()` would tick on a doc-only bump — a false ABI-change signal. The generator derives the **ABI-only** ordinal (#32).

**What stays here:** `release.sh` sets the 4-field **release version** in the snapshot's `interface.yaml` / metadata (its existing, legitimate job). The generator reads that and emits both ABI signals.

So this PR should shrink to *version-stamping only*, with the hash + version-emission deferred to linux_binder_idl#32. Once #32 lands (2.5.0) and the pin bump (#683 → binder 2.4.0, and later 2.5.0) is in, snapshots get correct, ABI-only `getInterfaceVersion()`/`getInterfaceHash()` for free.

cc: this came out of the review question on whether baking the hash in `release.sh` was the right place — it isn't; the generator is.

Ulrond:
Tracking note: the versioning design is consolidated in #633 (now *trustworthy frozen interface versioning*, with #658 folded in). Per that + linux_binder_idl#32, this PR re-scopes to **version-setting only** — the version-ordinal and contract-hash emission move to the generator (linux_binder_idl#32/#33).

Issue #662, opened 2026-06-30
Binder runtime: MW and vendor must ship separate bitness-matched binders; define kernel feature/version floor


Ulrond:
## Problem

The layer-aggregation model separates **MW** (`/mw`) from **vendor** (`/vendor`) as independently-mounted, dynamically-linked layers. Two facts collide:

1. **In-process linking cannot cross bitness.** A 32-bit process cannot load a 64-bit `.so` (ld.so: `wrong ELF class`). So a 32-bit MW **cannot in-process link** a 64-bit vendor userspace library. This is an ELF/ABI hard wall, not a build-flag issue.
2. **The kernel binder driver varies across the target range.** The userspace `libbinder` we ship (AOSP `android-13.0.0_r74`) requires binder-driver features that older kernels in the range may not implement (binderfs is 5.0+, freeze/extended-error ioctls are 5.x+). Docs currently state "Kernel 5.10+" (migration-guide) and "tested on 5.16.20" (linux_binder_idl) — **the actual supported floor is undefined**, and 4.9 has been mentioned as a target.

## Decision

- **MW builds and ships its own `libbinder`**, 32-bit (`TARGET_LIB32_VERSION=ON`; `CONFIG_ANDROID_BINDER_IPC_32BIT` per the 32-bit-userspace requirement).
- **Vendor builds and delivers its own binder requirement** (matched to the vendor's own bitness).
- Because the two bitnesses differ, **MW ↔ vendor communicate over the kernel binder driver (IPC)** — which bridges 32/64 — **not in-process**. Each layer owns a bitness-matched `libbinder`; neither links the other's.

## Open items / to define
- [ ] **Kernel feature/version floor** — audit which binder ioctls/features AOSP-13 libbinder uses unconditionally vs guards, and set the minimum supported kernel (audit in progress).
- [ ] **binderfs (5.0+) vs static `/dev/binder` node** — document provisioning for kernels below 5.0 if they're in scope.
- [ ] **Verify `CONFIG_ANDROID_BINDER_IPC_32BIT`** — BUILD.md prescribes it for "32-bit userspace on 64-bit kernel," but that flag is for 32-bit *kernels*; a 64-bit kernel normally serves 32-bit userspace via `compat_ioctl` without it. Confirm correct config.
- [ ] **servicemanager ownership** — which layer runs it; both sides use the same context.
[...]

PR #663, opened 2026-06-30
docs(standards): binder runtime layering, bitness & kernel requirements (#662)

Ulrond:
Part of #662.

Adds `docs/standards/binder_runtime_and_bitness.md` documenting the MW/vendor binder architecture:

- **Each layer ships its own bitness-matched libbinder** — MW 32-bit (`TARGET_LIB32_VERSION=ON`), vendor at its own bitness — and they communicate **over the kernel binder driver (IPC)**, which bridges 32/64. In-process linking can't cross ELF bitness, so the cross-bitness split is necessarily an IPC boundary.
- **Bitness rule** table (in-process vs cross-layer) and the kernel-bitness-independence note.
- **Kernel requirements** — binder config, device node (binderfs ≥5.0 vs static), 32-bit-userspace-on-64-bit-kernel compat path, stable protocol version.
- A placeholder for the **minimum-kernel feature matrix**, to be pinned from the libbinder feature audit (in progress under #662).

Wired into the mkdocs nav under Standards.

Follow-up commit will fill the feature matrix once the AOSP-13 libbinder ioctl audit completes.

### DIFF
--- a/docs/standards/binder_runtime_and_bitness.md
+++ b/docs/standards/binder_runtime_and_bitness.md
@@ -18,47 +18,70 @@ graph LR
     K -->|ioctl /dev/binder| D
 ```
 
-## Each layer ships its own bitness-matched libbinder
+## Each layer ships its own libbinder
 
-- **MW** builds and delivers its own `libbinder`/`libutils`, **32-bit** (`TARGET_LIB32_VERSION=ON`).
-- **Vendor** builds and delivers its own `libbinder`/`libutils`, matched to the vendor's own bitness (e.g. 64-bit).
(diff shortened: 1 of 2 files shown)


Ulrond:
Relocated to linux_binder_idl — this is a binder build/runtime topic (per-layer libbinder build, bitness, protocol version, kernel range), so it belongs with the binder toolchain, not rdk-halif-aidl standards docs. Folded into linux_binder_idl/BUILD.md via rdkcentral/linux_binder_idl#44 (issue rdkcentral/linux_binder_idl#43), on the 2.5.0 milestone. The mixed protocol-8 build gap is now tracked properly in rdkcentral/linux_binder_idl#42. Closing here.

Issue #672, opened 2026-07-03
Define per-layer build model: each layer builds its own HALIF + userspace binder; version alignment before AIDL V1


Ulrond:
## Context — how each layer builds today
Each layer (**MW**, **vendor**) builds and deploys **its own userspace binder** (see #662) **and its own HALIF interface libraries**. This is the model today, and it holds into the future — neither layer consumes the other's binder/HALIF binaries in-process; they interoperate over the kernel binder driver.

## Direction of travel (post AIDL V1)
Once interfaces are frozen at **AIDL V1**:
- **Vendor layer freezes** — it pins a specific frozen interface version per component and implements against it.
- **MW layer is free to choose** which version per component it consumes — it can mix versions across components.
- This works because the interface version + contract hash (getInterfaceVersion()/getInterfaceHash(), #633) let each side pin/negotiate independently, and each layer builds its own bitness/protocol-matched binder (#662).

## Short term (before AIDL V1) — we must ALIGN
Interfaces are **not frozen yet** (V1 not reached), so mixing versions isn't safe. In the interim:
- **MW and vendor must align** on the same interface version per component (no per-component version divergence yet).
- But we must **correctly deploy the build requirements for each layer to build its own userspace binder** — the per-layer, bitness/protocol-matched binder build (#662), plus the HALIF interface source/libs each layer compiles.

## What this ticket needs
1. **Document the per-layer build model**: each layer builds its own HALIF interface libraries + its own userspace binder (bitness/protocol matched — #662), and the build inputs each layer requires.
2. **Define the versioning direction of travel**: pre-V1 alignment → post-V1 vendor-freeze / MW-free-choice per component.
3. **Deploy the build requirements per layer** so each layer can build its own userspace binder correctly (the interim alignment + the eventual independence).

## Related
- #662 (per-layer binder: bitness/protocol, kernel floor)
- #633 (interface version + contract hash — enables the post-V1 independence)
- #671 (layer-aggregation filesystem test/demo)
- `vsi/filesystem/current/docs/directory_and_dynamic_linking_specification.md`

Ulrond:
Raised #807 for the layout half of item 3 here.

[...]

PR #673, opened 2026-07-03
test/demo: emulate layer-aggregation filesystem + cross-layer dynamic linking (#671)

### LINKED ISSUE #671: test/demo: emulate the layer-aggregation filesystem (MW/vendor) + cross-layer dynamic linking
## Goal
Our Yocto test (`tests/fake-yocto/`) stages the SDK into a *flat* sysroot. But the real deployment uses the **layer-aggregation architecture** ([spec](../blob/develop/vsi/filesystem/current/docs/directory_and_dynamic_linking_specification.md)): independently-mounted read-only layers (`/mw`, `/vendor`, `/apps`), each with aggregation directories (`lib/`, `ld.so.conf.d/`, `bin/`, `systemd/`…) of symlinks to module resources, wired for dynamic linking via `/etc/ld.so.conf.d/<layer>.conf` + `ldconfig`.

We should model that in our tests **and** a runnable demo — so we validate that:
- the MW layer (binder SDK + HAL interface libs) and a vendor layer stage into the aggregation structure correctly, and
- a consumer resolves libraries **across layers** through the `ld.so.conf.d` aggregation (not rpath / not a flat sysroot).

## Scope
- `tests/fake-layers/run-fake-layers.sh` — offline emulation (no root/mount, like fake-yocto):
  - build the binder SDK + HAL libs, stage them as **MW-layer modules** (`/mw/binder`, `/mw/halif`) with aggregation symlinks (`/mw/lib/* -> ../<module>/lib/*`) + `/mw/ld.so.conf.d/`,
  - stage a small **vendor-layer** consumer module,
  - wire `/etc/ld.so.conf.d/{mw,vendor}.conf` → run `ldconfig` → assert the HAL/binder libs are registered and a cross-layer consumer resolves its deps via the aggregated cache.
  - Doubles as the **demo** (prints the layer tree; `--keep` to inspect).
- Reference from `tests/README.md`; link the spec.

Ulrond:
Closes #671.

Our Yocto test stages the SDK into a **flat sysroot**. Real deployment uses the **layer-aggregation** model ([spec](../blob/develop/vsi/filesystem/current/docs/directory_and_dynamic_linking_specification.md)): independently-mounted `/mw` and `/vendor` layers, each exposing modules via aggregation dirs (`lib/`, `ld.so.conf.d/`) of symlinks, wired for dynamic linking through `/etc/ld.so.conf.d/<layer>.conf` + `ldconfig`.

## What it does
`tests/fake-layers/run-fake-layers.sh`:
- stages the binder SDK + HAL interface libs as **MW-layer modules** (`/mw/binder`, `/mw/halif`) with aggregation symlinks + per-module `ld.so.conf.d`,
- builds a **vendor-layer** library that links a HAL lib **from the MW layer** (the cross-layer dependency),
- wires `/etc/ld.so.conf.d/{mw,vendor}.conf` and runs `ldconfig`,
- asserts: aggregation symlinks resolve; `ldconfig` registers the binder+HAL libs via the aggregated `ld.so.conf.d`; and the vendor lib's MW-provided dependency **resolves across layers** (nothing "not found").

Doubles as a **demo** (prints the emulated layer tree; `--keep` to inspect). No root/mount needed; **skips cleanly** without `ldconfig`/`ldd`/compiler.

## Verified
4/4 assertions pass locally (aggregation symlinks, `ldconfig` registration, cross-layer resolution of `libcommon` from `/vendor` → `/mw`).

Relates to #672 (per-layer build model) and #662 (per-layer binder).

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -62,8 +62,11 @@ for dynamic linking via `/etc/ld.so.conf.d/<layer>.conf` + `ldconfig`:
 
 Where `fake-yocto/` proves the flat-sysroot build, this proves the **layered**
 runtime: the MW layer provides the binder runtime + HAL interface libs, a
-vendor-layer library links a HAL lib from the MW layer, and the test asserts
-the cross-layer dependency resolves **through the `ld.so.conf.d` aggregation**
-(not rpath, not a flat sysroot). No root/mount needed — layers are emulated in
-a work dir and validated with `ldconfig` (registration) + `ldd` (resolution).
-It doubles as a demo (prints the emulated layer tree; `--keep` to inspect).
+vendor-layer library links a HAL lib from the MW layer, and the test asserts the
+cross-layer dependency via **layer aggregation** — not rpath, not a flat sysroot.
+Two checks: `ldconfig` **registers** the aggregated libs through the per-layer
+`ld.so.conf.d` entries (verified in the cache), and `ldd` **resolves** the
+vendor→MW dependency against the aggregated `lib/` dirs (host `ldd` can't read
+the work-dir cache, so resolution is checked with `LD_LIBRARY_PATH` pointed at
+those dirs). No root/mount needed — layers are emulated in a work dir. It doubles
+as a demo (prints the emulated layer tree; `--keep` to inspect).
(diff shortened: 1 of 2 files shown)


Ulrond:
Copilot review addressed: vendor lib now links via the MW aggregation dir (`${MW}/lib`), and the README now distinguishes ldconfig *registration* (via `ld.so.conf.d`) from ldd *resolution* (via `LD_LIBRARY_PATH` against the aggregated dirs). On the top-level `<layer>.conf` pointing at `<layer>/lib`: that dir aggregates every module's libs as symlinks, so it's a valid aggregation model (the per-module confs are also symlinked into `ld.so.conf.d/`).

Issue #658, opened 2026-06-30
feat(release): enforce version bump matches AIDL change class (additive-within-generation gate)


Ulrond:
## Problem
The runtime interface-compatibility check defined in #633 — `generation(server)==generation(client) && server>=client` — is only **trustworthy** if our versioning discipline actually holds:

- `minor`/`patch` bumps (`0.X.Z.W`) are **purely additive** (append methods/fields only), and
- every **breaking** change forces a **generation** bump (`0.<gen>` increments).

Nothing enforces this mechanically today. A breaking change could land under a minor bump, and a consumer relying on `server >= client` would then be told "compatible" when it isn't. Until this gate exists, `getInterfaceVersion()` comparison is **advisory, not guaranteed**.

## Proposed solution
At release/freeze time (`release.sh create_snapshot`), for each component:

1. Produce a canonical surface dump of the **previous released** AIDL and the **about-to-be-frozen** AIDL.
2. Classify the structural diff as **additive** (added methods/fields/enum values at the end), **breaking** (removed/renamed/reordered/retyped), or **none**.
3. Assert the planned version bump matches:
   - breaking ⇒ **generation** must increment;
   - additive ⇒ **minor** (or patch) bump, generation unchanged;
   - none ⇒ no snapshot.
4. Fail the release with a clear diff if the bump and the change class disagree.

## Mechanism
Built on **linux_binder_idl #27** (`dump-surface` / `diff-surface`) — that toolchain change provides the canonical dump + structural classification; this issue is the **rdk-halif-aidl release-side gate** that consumes it. Gated on #27 landing.

## Related
- #633 — interface version + contract hash scheme (defines the compatibility predicate this gate protects).
- linux_binder_idl#27 — `dump-surface`/`diff-surface` subcommands (the classification engine).

Ulrond:
Folded into #633 (retitled *trustworthy frozen interface versioning*). #658's change-class enforcement gate and #633's version/hash baking are two halves of one capability — *produce* the version/hash and *guarantee* it's honest — so they're now a single tracked design, reconciled to the ABI-only encoding decided in linux_binder_idl#32. The gate mechanism (dump-surface/diff-surface via linux_binder_idl#27) is captured in #633.

PR #691, opened 2026-07-06
fix(release): getInterfaceVersion() reports the release version as positional int (#633)

Ulrond:
Refs #633. Corrects #690's version scheme to the decided design.

## What a client sees
| State | `getInterfaceVersion()` | `getInterfaceHash()` |
|---|---|---|
| Frozen 0.2.0.0 | `20000` (decodes to 0.2.0.0 on sight) | real hash |
| Frozen 0.3.0.0 | `30000` | real hash |
| Frozen 1.0.0.0 | `1000000` | real hash |
| Dev tree (`current/`) | unchanged | `"notfrozen"` ← pre-freeze marker |

- **Version = the release version itself**, fixed-width positional int (two digits per field): self-describing, no lookup table needed — the requirement for pre-AIDL-versioning reporting. Decode: `prefix=(v/1000000)%100, generation=(v/10000)%100, minor=(v/100)%100, patch=v%100`.
- **Monotonic forever**: the same scheme covers 0.x and 1.x+ (`20000 < 30000 < 1000000`), so `server >= client` ordering never breaks — no scheme switch, ever.
- **Collision-safe**: two-digit fields mean `0.10.0.0 → 100000` ≠ `1.0.0.0 → 1000000`; malformed or >99 fields leave the snapshot unfrozen rather than emit a wrong number.
- **Hash unchanged from #690**: toolchain `aidl_hash_gen`, labelled with the emitted version int (hash binds content + version).

## Verification
- Encoding: `0.1.0.0→10000, 0.2.0.0→20000, 0.1.0.1→10001, 1.2.3.4→1020304, 0.10.0.0→100000≠1.0.0.0→1000000`; `0.2.0`/`0.100.0.0`/garbage → unfrozen.
- Hash deterministic, version-bound (label 20000 vs 30000 → different digests).
- `interface.yaml` stamp round-trip OK; `current/` restored on every exit path; `bash -n` clean.

The change-class enforcement gate (was #658) remains open scope on #633, pending linux_binder_idl#27.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1783,37 +1783,44 @@ is_buildable_component() {
 # Frozen interface VERSION + contract HASH (#633)
 # ----------------------------------------------------------------------------
 # getInterfaceVersion() reports the RELEASE version itself, encoded as a
-# fixed-width positional int32 — self-describing, no lookup table: each
-# X.Y.Z.W field is zero-padded to two digits and concatenated, so
-#   0.2.0.0 -> 00 02 00 00 ->   20000
-#   0.3.0.0 -> 00 03 00 00 ->   30000
-#   0.1.0.1 -> 00 01 00 01 ->   10001
-#   1.0.0.0 -> 01 00 00 00 -> 1000000
-# Decode: prefix=(v/1000000)%100 generation=(v/10000)%100 minor=(v/100)%100
-# patch=v%100. Monotonic across ALL releases (0.x < 1.x < ...) because the
-# same scheme is used forever — there is no later switch to bare ordinals.
-# Max 99.99.99.99 = 99,999,999, well under int32 max; a field >= 100 is not
-# encodable and leaves the snapshot unfrozen rather than emit a wrong number.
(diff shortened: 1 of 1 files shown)


Ulrond:
Updated to the decided **1-2-2-1** field widths (`era | major | minor | doc`):

| Release | `getInterfaceVersion()` |
|---|---|
| 0.1.0.0 | `1000` |
| 0.2.0.0 | `2000` |
| 0.1.0.1 | `1001` |
| 0.10.0.0 | `10000` |
| 0.99.99.9 | `99999` (pre-era max) |
| 1.0.0.0 | `100000` |

Decode: pad to 6 digits, read `1|2|2|1`. Same scheme across eras — monotonic forever, verified strictly increasing across a mixed sample. Out-of-width fields (era/doc > 9, major/minor > 99) leave the snapshot unfrozen; **the doc field caps at 9, so a 10th doc-only respin of a minor forces a minor bump** (accepted trade-off).

Issue #566, opened 2026-06-04
Stop committing generated .h/.cpp under current/; regenerate at build time and commit only via release.sh into <version>/


Ulrond:
## Problem

`<module>/current/include/` and `<module>/current/src/` are toolchain-generated `.h`/`.cpp` files that we commit as snapshots in develop. That conflates two distinct things:

1. **AIDL source** — authored, must be in git
2. **AIDL output** — derivative, regenerable from the source by the toolchain

The conflation causes a class of problems we've been firefighting:

- **Drift** (#564) — committed generated headers fall out of sync with the AIDL surface (just fixed in #565 via a smoke-test drift check, but only as an interim).
- **Cascade-commit discipline** — every PR that changes one module's AIDL must regenerate and re-commit every downstream module's `.h`/`.cpp`. Documented in versioning-sop.md but always fragile.
- **PR diff noise** — 10 AIDL lines produce ~200 lines of regenerated bindings in the diff. Reviewers lose focus on the actual surface change.
- **Repo size** — generated code roughly doubles `current/`.
- **Incremental-build hazards** — stale committed `.h` + freshly-regenerated `.cpp` mid-rebuild causes cryptic "no declaration matches" errors (e.g. the #542 build failure).

The root-cause fix is to stop committing generated code in `current/` entirely.

## Why `current/` only — and why `<version>/` snapshots must still be committed

Frozen `<module>/<version>/` snapshot directories ARE the release artifact — consumers pin to `common@0.1.0.0`, not to `common@current`. The `<version>/include/` and `<version>/src/` must be in git because they're the immutable surface a downstream HAL implementation builds against.

But **nobody pins to `current/include/`**. `current/` exists only for in-tree development. Anyone touching `current/` is a developer with the toolchain installed and runs the build before opening a file. There's no consumer who clones the repo and reads `current/include/` directly — they read `<version>/include/`.

So the asymmetric fix:

- `current/include/` and `current/src/` — **regenerated at build time, never committed**. `.gitignore`d.
- `<version>/include/` and `<version>/src/` — **committed**. Created by `./release.sh` during cohort freeze, written once, immutable.

### 1. Delete committed `current/include/` + `current/src/` content

One cleanup commit removing every existing `*/current/include/` and `*/current/src/` directory's contents. Authored AIDL under `*/current/com/...` stays.

### 2. `.gitignore` `*/current/include/` and `*/current/src/`

```
*/current/include/
*/current/src/
```

[...]

PR #729, opened 2026-07-10
refactor(examples): drop committed generated C++ from aidl_versioning demo (#660)

### LINKED ISSUE #660: build: aidl_versioning example commits generated C++ and isn't on the module-local model — remove + rewire
## Summary

The `examples/aidl_versioning/` demo commits **296 generated C++ files** under `stable/generated/`, and its build consumes that committed tree. This is inconsistent with the post-#566 model, where generated C++ for `current/` is gitignored and regenerated by the toolchain and the central `stable/` tree is retired. The committed generated C++ should be removed **and** the example rewired to regenerate on build.

This is **standalone, demo-only** work — `examples/aidl_versioning/` is built manually and is not referenced by the root CMake, `build_modules.sh`, or any CI workflow, so nothing in the main build/CI is affected.

## Why it can't be a simple delete

The example's per-version component CMakeLists hard-code include paths into the committed tree, e.g.:

- `examples/aidl_versioning/client/car_client_ver_1/CMakeLists.txt`
  `target_include_directories(... ${INTERFACES_ROOT_DIRS}/stable/generated/car/include/)`
- `examples/aidl_versioning/server/car_service_ver_2/CMakeLists.txt` (same pattern)
- also `car_client_ver_3`, `car_service_ver_3`

So removing `stable/generated/` without rewiring breaks the demo build. The two steps must land together.

## Proposed fix

Rewire the example to the module-local model so it regenerates its C++ on build (mirroring `build_modules.sh` / `target_build_interfaces_libraries()`) instead of consuming a committed `stable/generated/` tree. Either:

1. Generate per-version C++ into module-local `{include,src}` (gitignored) during the example's configure step and point the CMakeLists there, then `git rm` the committed `stable/generated/` and gitignore the path; or
2. Retire the `stable/`-based example layout entirely and re-model the versioning demo on the current component layout.

Source AIDL under `examples/aidl_versioning/stable/versioned_aidl/` should be retained either way.

## Acceptance

- No generated C++ committed under `examples/aidl_versioning/`.
- `examples/aidl_versioning/` builds from a clean checkout (regenerating its C++).

## Notes

Out of scope for PR #387 (which is scoped to the CryptoEngine/KeyVault HALs). #387 briefly removed these files then reverted that change to keep its scope clean; this ticket tracks doing it properly as standalone work.

Ulrond:
Closes #660.

## What
The `examples/aidl_versioning/` demo committed **296 generated C++ files** under `stable/generated/`, and its server/client CMakeLists pointed `target_include_directories` at that tree. That is off the post-#566 model, where generated bindings are toolchain output regenerated on build, not committed source.

## Changes
- **`git rm` the committed `stable/generated/` tree** (296 files).
- **Remove the redundant `target_include_directories(... stable/generated ...)`** from all four server/client targets. `target_link_interfaces_libraries(<tgt> car-vN-cpp)` links the interface library **PUBLIC**, and that generated target already exposes its freshly-generated headers **PUBLIC** (`build-tools/linux_binder_idl/host/CMakeLists.inc:318`), so the explicit include was redundant — and latently **wrong**: all four pointed at the *same* committed `car/include/` regardless of the v1/v2/v3 they actually link.
- **`.gitignore`** `examples/aidl_versioning/stable/generated/` so it cannot return.
- Refresh the stale "generated sources" comments in the server/client CMakeLists.

Generation remains driven from `stable/versioned_aidl/` (kept — that's the demo's versioned-AIDL source of truth).

## Verification
- Confirmed by inspection of the toolchain macro that generated headers propagate transitively (`PUBLIC` include on the aidl target + `PUBLIC` link to the consumer) — the removed explicit includes were redundant.
- Configured the demo against the repo-root toolchain: it loads and **reaches the generation stage reading the example's `interfaces/`/`versioned_aidl/`**, confirming generation is source-driven, not dependent on the removed C++.
- **Out of scope / pre-existing:** a full one-step demo build still needs the toolchain dependency-tree bootstrap (`dependencies.txt`), a gap in the demo's manual build that predates and is independent of this cleanup (it scans `interface.yaml`, not the generated C++). Flagging rather than expanding scope here.

## Risk
Low — demo-only, not referenced by the root CMake, `build_modules.sh`, or any CI workflow.


Ulrond:
Closing in favour of redoing this as option 2 of #660 — re-model the example on the current component layout — rather than rebasing this branch.

Not because the work is wrong. It found something worth keeping, and #660 now records it: all four server/client targets pointed `target_include_directories` at the *same* committed `car/include/` regardless of which of v1/v2/v3 they actually linked. In a demo whose entire subject is versioning, every target was compiling against one version's headers. That is a bug the deletion happened to expose, and it should not be lost.

Why redo rather than rebase:

- The branch is `BEHIND` and last touched 2026-07-10. It is a 15,236-line deletion against a tree that has moved since, and it predates the module-local rework it is trying to align with.
- Its own verification says a full one-step demo build still needs the toolchain dependency-tree bootstrap, so the demo would not build after merging. That gap is the actual acceptance criterion in #660 — "builds from a clean checkout" — and it stays open either way.
- Separately, `develop` carries two dangling symlinks in the example: `install_binder.sh -> ../../install_binder.sh` and `targets -> ../../targets/`, neither of which exists at the repo root. So the demo is broken independently of the committed C++, and a deletion-only change cannot fix it.

The branch stays available if any of it is worth cherry-picking.

PR #735, opened 2026-07-13
rag: present sensor as sensor/motion + sensor/thermal

### LINKED ISSUE #734: rag: present sensor as sensor/motion + sensor/thermal (report-only)
The `sensor` module ships two cleanly separated sub-interfaces — `com.rdk.hal.sensor.motion` and `com.rdk.hal.sensor.thermal`, each with its own HFP YAML — but the RAG report shows a single `sensor` row. The programme tracks them as two (component:motion_sensor / component:thermal_sensor; ENT-8650 vs ENT-8651).

Make `generate_rag_report.sh` render `sensor` as two rows (`sensor/motion`, `sensor/thermal`) so the dashboard reflects both capabilities. Report-only: the module stays one (one metadata.yaml, one interface.yaml, one version); both rows share its status/version/reviews. No interface or versioning change.

Ulrond:
Closes #734.

The `sensor` module ships two cleanly separated sub-interfaces — `com.rdk.hal.sensor.motion` and `com.rdk.hal.sensor.thermal`, each with its own HFP YAML — but the RAG report showed a single `sensor` row. This expands it into two rows so the dashboard reflects both capabilities, matching the programme's two-component view (`component:motion_sensor` / `component:thermal_sensor`; ENT-8650 vs ENT-8651).

**Report-only — no interface or versioning change.** `sensor` stays one module (one `metadata.yaml`, one `interface.yaml`, one version); both rows share its status/version/reviews. The generator now iterates a per-component path list (just the component path, except `sensor` → `motion` + `thermal`), with the counters moved inside that loop so rows and tallies stay consistent.

| | Before | After |
|---|---|---|
| Components | 31 | 32 |
| GREEN | 17 | 18 |
| sensor rows | `sensor` | `sensor/motion`, `sensor/thermal` |

**Incidental:** the regenerated report also refreshes three versions that were stale in the committed copy (avclock 0.2.0.1, common 0.2.1.0, videodecoder 0.2.0.1) — pre-existing drift from metadata bumps without a report regen.

If independent versioning/lifecycle per sensor type is wanted later, that's the heavier module-split option (deferred).

### DIFF
--- a/RAG_STATUS_REPORT.md
+++ b/RAG_STATUS_REPORT.md
@@ -2,9 +2,9 @@
 
 | | |
 |---|---|
-| **Generated** | 2026-07-07 |
-| **Components** | 31 |
-| 🟢 **GREEN** | 17 |
+| **Generated** | 2026-07-13 |
+| **Components** | 32 |
+| 🟢 **GREEN** | 18 |
 | 🟡 **AMBER** | 14 |
 | 🔴 **RED** | 0 |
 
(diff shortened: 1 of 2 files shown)


Ulrond:
Redundant — the RAG sensor fix landed on release/0.22.0 via #736 and reaches develop through the release→develop back-merge, so a separate develop PR isn't needed.

PR #149, opened 2025-07-10
Add #105 - Split Codec to Codec / Profile and upgraded documentation

Ulrond:
Split Codec / Profile out of capabilities and corrected configuration

PR #740, opened 2026-07-16
feat(audiodecoder): add bitrateBps to FrameMetadata (#139)

### LINKED ISSUE #139: Task: Audio Decoder bitrate review
### Goal

bitrateBps is missing from `FrameMediadata.aidl` , should it be added?

```yaml
parcelable FrameMetadata {
    ...
    int bitrateBps;  // Optional: bitrate in bits per second
    ...
}
```

### Notes (Optional)

_No response_

Ulrond:
Closes #139.

Adds `bitrateBps` to `audiodecoder` `FrameMetadata` — the encoded bitrate of the source elementary stream for the frame, in bits per second.

- **Informational only**: metrics, UI reporting, stream diagnostics. Does not affect decoding or presentation.
- `0` when the decoder reports no bitrate (raw PCM, or a codec/container carrying none). For VBR streams it is the instantaneous/most-recent bitrate derived for the frame.
- `int` (not `long`): the maximum lossless audio bitrate (Dolby TrueHD, DTS-HD MA) is well under 2^31 bps.

**Change class:** Minor Change — additive field on the shipped `audiodecoder` parcelable (0.2.0.0 → next). Appended at the **bottom**, immediately before the `ParcelableHolder extension`, so the field ordinals of the released 0.2.0.0 snapshot are preserved.

### DIFF
--- a/audiodecoder/current/com/rdk/hal/audiodecoder/FrameMetadata.aidl
+++ b/audiodecoder/current/com/rdk/hal/audiodecoder/FrameMetadata.aidl
@@ -133,6 +133,24 @@ parcelable FrameMetadata {
 	byte[] SoCPrivate;
 
 	/**
+	 * Bitrate of the source elementary stream for this frame, in bits per second.
+	 *
+	 * The encoded bitrate reported by the decoder for the source audio
+	 * (e.g. ~640000 for 640 kbps AC-3, ~1500000 for a DTS core). For variable
+	 * bitrate streams this is the instantaneous/most-recent bitrate the decoder
+	 * derives for this frame.
+	 *
+	 * Informational only — used for metrics, UI reporting, and stream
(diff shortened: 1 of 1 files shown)


Ulrond:
Good catch — Copilot is right. Inserting `bitrateBps` after `sourceCodec` shifted the ordinals of every following field, breaking wire compatibility with the released `0.2.0.0` parcelable.

Fixed in `09e6e3c4`: `bitrateBps` is now appended at the bottom, immediately before the final `ParcelableHolder extension` (which stays last), so no existing field's position moves.

Ulrond:
Closing unmerged. The stream bitrate is derivable by the client from the buffers it already submits, and the capability ceiling belongs in `CodecSupport` / the HFP (#149, #768). Full write-up: https://github.com/rdkcentral/rdk-halif-aidl/issues/139#issuecomment-5632348252

PR #756, opened 2026-07-27
feat(metrics): add com.rdk.hal.metrics — device metrics HAL (#754)

### LINKED ISSUE #754: Task: Add metrics provider HAL (com.rdk.hal.metrics) — per-instance KVP metric sources
## Motivation

Playback-quality metrics are a certification requirement across partners: frames dropped/repeated, decode errors, underflow episodes with durations, AV-sync excursions — with defined freshness and atomicity. Today each SoC exposes these ad-hoc (debug procfs, per-element properties, or not at all), so middleware cannot read them portably and cannot meet the accuracy contracts the certifications state.

## Proposal

Add **`com.rdk.hal.metrics`** — a general device metrics interface, not an A/V one. It carries named numeric values from whoever measures them to whoever consumes them, organised into **domains**. A/V playback is the `av` domain; `cpu` and `memory` are domains a platform may add later without touching it or the interface.

```text
IMetricsManager.aidl              — catalog + live source enumeration
IMetricsSource.aidl               — one per source: getAll / getFieldsByName / getField / setField
IMetricsManagerEventListener.aidl — sources appearing/disappearing
MetricKVPair.aidl · MetricFieldInfo.aidl
MetricElementInfo.aidl · MetricDomainInfo.aidl · Capabilities.aidl
```

### Naming

Every metric name is **four segments, fully qualified, no short form**:

```text
<domain>.<element>.<instance>.<field>

av.video_decoder.0.frames_decoded
av.video_sink.1.frames_dropped_late
av.clock.0.sync_offset_ms
cpu.core.3.utilisation_pct
```

The first three address a source (`getSource("av.video_decoder.0")`); the fourth selects a field within it. **The path is the identity** — there is no source-id parcelable, because modelling it twice only creates a way for the two to disagree.

Names are by **subject, not producer**. Which block sources a figure differs per SoC, and for some fields it is middleware rather than the SoC at all, so a producer-shaped name would move between products for the same metric.

### Contract points

- **No enums, no value union.** Every key is a declared string; **every value is int64** (AIDL `long`). A metric is a count, a duration, a byte figure or an offset — all signed 64-bit. Counters use the positive range; a signed field like `sync_offset_ms` uses the sign.
- **The source is the atomicity boundary.** One `getAll()` is one coherent snapshot, so paired fields (frames decoded vs presented) can never be read torn. This is an obligation on the implementation, not a property to be discovered — a source spanning two hardware blocks latches both.
- **Counters monotonic** since source creation, no reset on flush or seek; consumers compute deltas.
- **Declared freshness per element**, platform floor 50 ms. An element may declare tighter, never looser.
- **Absence is absence.** A field a product cannot measure is left undeclared and omitted from reads — never served as 0. "Cannot measure it" and "measured zero" must stay distinct.
- **No SoC-private namespace.** Every name is defined in its domain dictionary, whatever produces it, so no consumer ever grows per-SoC code. A figure only one SoC can produce still gets one dictionary entry and simply goes undeclared elsewhere.
- **Episodic conditions are fields, not events.** Each is reported in two parts: a `counter` totalling occurrences (`underflowed`, `decode_errors`, `silence_event_count`) and `current` fields describing the most recent one (`last_underflow_duration_ms`, `last_decode_error_pts_ms`, `last_decode_error_reason`). A counter that advanced between polls makes the occurrence visible; the `last_*` fields make it diagnosable. Both arrive in the same snapshot, so an occurrence and the counters around it are always consistent. A PTS the SoC cannot derive is left **undeclared**, not sent as `-1`.
- **No per-caller state.** Every read is a snapshot of what the source holds now, so any consumer reads at any cadence without affecting another. Fan-out is a middleware concern.

### Declaration

The metric set is **not part of the ABI**. Adding a field, an element or a domain is a declaration change: no interface freeze, no version bump, no coordinated consumer rebuild — which is what lets the metric set track the partner certification set, which moves annually, without the HAL moving with it.

`metrics/current/hfp-metrics.yaml` declares, per element: its fields (`name`, `unit`, `kind`, `writable`), `instances` (how many the hardware supports — the live set comes from `getSourcePaths()`) and `pollCadenceMs`.

Two conventions the file enforces:

- **Declare the whole dictionary, comment out what you cannot serve**, one line each with `# NOT SUPPORTED — <reason>`. The file then reads as a checklist, so a reviewer sees what the SoC does not do rather than having to notice something missing.
- **`instances` is the ceiling, not the live count.** It makes dual-decode capability checkable before boot, lets a test assert no source index >= `instances` appears, and must agree with the owning HAL's own HFP — a CI-catchable cross-check.

### `av` domain elements

| Element | Carries |
|---|---|
| `video_decoder` | frames decoded/dropped, decode errors, decode latency |
| `video_sink` | frames received/presented, late and FRC drops, repeats, freeze, underflow, render latency, buffer depth |
| `audio_decoder` | audio decode counters and errors |
| `audio_sink` | audio presentation, underflow, digital silence, buffer depth |
| `mixer` | mix-path counters where the platform mixes concurrent sessions |
| `clock` | A/V sync offset, max absolute offset, time over threshold, resync count and magnitude |

Derived domains (CMCD and similar, computed as a stated function of measured fields) never cross this interface: a vendor cannot declare one and never sees one. They are computed by middleware from its own state.

## Acceptance criteria

- AIDL package `com.rdk.hal.metrics` under `metrics/current/`, standard HAL documentation methodology, all types `@VintfStability`.
- `metrics/current/hfp-metrics.yaml` with its pykwalify schema, and the normative field dictionary for the `av` domain (name, unit, kind, semantics, population rule per field).
- `versions.yaml` + `CHANGELOG.md` entries; L1/L2 test skeletons consistent with the other modules.

---

_Design updated 2026-07-27. The original proposal used an int-backed `MetricKey` enum, a `MetricSourceId` parcelable of component-kind + instance, and `getMetricsMulti(keys[])`. Those are superseded: keys are declared strings in a four-segment fully-qualified path, the path is the identity, and reads are per source. See the comment below for the full delta._

Ulrond:
Closes #754.

Adds **`com.rdk.hal.metrics`** — a general device metrics interface, not an A/V one. It carries named numeric values from whoever measures them to whoever consumes them, organised into **domains**. A/V playback is the `av` domain; `cpu` and `memory` are domains a platform may add later without touching it or this interface.

## Naming

Four segments, fully qualified, no short form:

```text
<domain>.<element>.<instance>.<field>

av.video_decoder.0.frames_decoded
av.video_sink.1.frames_dropped_late
av.clock.0.sync_offset_ms
cpu.core.3.utilisation_pct
```

The first three address a **source** (`getSource("av.video_decoder.0")`); the fourth selects a field within it. **The path is the identity** — there is no source-id parcelable, because modelling it a second time only creates a way for the two to disagree.

Names are by **subject, not producer**. Which block sources a figure differs per SoC, and for some fields it is middleware rather than the SoC at all, so a producer-shaped name would move between products for the same metric.

## Three properties

- **Every name is a declared string.** No key enums, no compiled-in metric list, no value union.
- **Every value is int64** (AIDL `long`). Counters use the positive range and never the sign; a signed field such as `sync_offset_ms` uses it.
- **The metric set is not part of the ABI.** Adding a field, an element or a whole domain is a declaration change — no interface freeze, no version bump, no coordinated consumer rebuild. That is what lets the metric set track the partner certification set, which moves annually, without the HAL moving with it.

## What is in the change

```text
metrics/metadata.yaml
metrics/README.md
metrics/current/interface.yaml · CMakeLists.txt · mkdocs.yml · docs/metrics.md
metrics/current/hfp-metrics.yaml · hfp-metrics-schema.yaml
metrics/current/com/rdk/hal/metrics/
    IMetricsManager.aidl              catalog + live source enumeration
    IMetricsSource.aidl               getAll / getFieldsByName / getField / setField
    IMetricsManagerEventListener.aidl sources appearing and disappearing
    Capabilities.aidl · MetricDomainInfo.aidl · MetricElementInfo.aidl
    MetricFieldInfo.aidl · MetricKVPair.aidl
```

Registered in `versions_current.yaml` and the docs nav.

## Points worth reviewing

**The source is the atomicity boundary.** One `getAll()` is one coherent snapshot, so paired counters (frames decoded vs presented) cannot be read torn. This is an obligation on the implementation rather than a property to be discovered — a source spanning two hardware blocks latches both.

**Absence is absence.** A field a product cannot measure is left undeclared and omitted from reads, never served as `0`. A PTS that cannot be derived is omitted, not sent as `-1`. "This SoC cannot measure it" and "it measured zero" are different facts and the interface keeps them different.

**No SoC-private namespace.** Every name is defined in its domain dictionary, whatever produces it, so no consumer ever grows per-SoC code. A figure only one SoC can produce still gets a dictionary entry and simply goes undeclared elsewhere. This is the main thing to push back on if you disagree — it means a vendor cannot add a private counter without a dictionary entry, so **dictionary admission is on the critical path for vendor adoption** and needs a named owner and a turnaround commitment.

**Episodic conditions are fields, not events.** Each is reported in two parts — a `counter` totalling occurrences (`underflowed`, `decode_errors`, `silence_event_count`) and `current` fields describing the most recent one (`last_underflow_duration_ms`, `last_decode_error_pts_ms`, `last_decode_error_reason`, `last_decode_error_vendor_code`). A counter that advanced between polls makes the occurrence visible; the `last_*` fields make it diagnosable. Both arrive in the same `getAll()` snapshot, so an occurrence and the counters around it are always mutually consistent.

**The HAL holds no per-caller state.** Every read is a snapshot of what the source holds now, so any consumer reads at any cadence without affecting another; fan-out is a middleware concern. The trade against an event buffer is burst fidelity — several occurrences inside one poll interval advance the counter by several and leave `last_*` describing the newest only. Rates and totals stay exact, and no consumer requirement asks for the middle of a burst.

**`instances` is the ceiling, not the live count.** Declaring it makes capability checkable before the product boots, lets a test assert no source index `>= instances` ever appears, and gives CI a cross-check against the owning HAL's own feature profile. The live set comes from `getSourcePaths()` because it is dynamic — a picture-in-picture decoder exists only while the second session does.

## Departure from the per-HAL house style

`videodecoder` and its peers use an int-backed `Property` enum with a `PropertyValue` union — correct for a fixed control surface of heterogeneous settings that changes only when the HAL changes. Metrics is an open set whose membership differs per product and per SoC generation, and whose values are uniformly numeric. So the key is a string path, the value is a bare `long`, and the set of both is declared data.

## Still to come

- Normative field dictionary for the `av` domain (name, unit, kind, semantics, population rule per field) — drafted middleware-side, to be contributed here.
- L1/L2 test skeletons consistent with the other modules.


Ulrond:
Design delta versus the original #754 proposal, for anyone who reviewed that version:

| #754 as filed | This PR |
|---|---|
| `MetricKey.aidl` — int-backed key enum | No enums; names are declared strings |
| `MetricSourceId` (component kind + instance) | Dropped — the path is the identity |
| `getMetricsMulti(keys[])` | `getAll()` / `getFieldsByName()` on a source |
| Keys per AV component kind | `<domain>.<element>.<instance>.<field>`, with `cpu`/`memory` on the same contract |
| — | int64 only, no `PropertyValue` union |
| — | Declaration file + schema, `instances`, absence-not-zero |
| Ring with sequence gaps | Buffer sized to one poll interval, single reader, loss counted |

Two things from the original that were right and are unchanged: **one reader** (middleware fans out; the HAL holds no cursor state) and **atomic multi-read**.

outdooruseonly:
For metrics.
The interface MUST have the detailed description of each metric and event FULLY documented. It is not complete without it.
For me this is the most important thing.
If the descriptions are not clear the VTS will be difficult and every implementation will vary.

Ulrond:
@outdooruseonly — agreed, and this is now in the branch.

Every field carries a normative description saying what it counts and when it moves, authored in `hfp-metrics.yaml` alongside its unit, kind and provider. `docs/vendor_field_dictionary.md` is the readable table generated from it — 45 fields across `video_decoder`, `audio_decoder`, `audio_sink` and `clock`. A sample entry:

| Field | Type · unit · kind | Provider | Definition and population rule |
|---|---|---|---|
| `underflowed` | int64 · episodes · `counter` | **Driver** | Count of video buffer-underflow **episodes**. +1 at each episode **start** (no frame to present at render time); ends when a new frame is presented. |

On your VTS point specifically — a prose description alone still leaves room to vary, so the parts a test needs to assert are stated as contract rather than left to the wording:

- **`kind` fixes how a value may be read.** `counter` is cumulative since source creation, monotonically non-decreasing, never reset on flush or seek. `current` is a live sample, absolute, never summed. `high_water` is a monotone maximum. `config` is the present value of a tunable. A test can assert those properties without reading English.
- **A field with no description fails the build.** `scripts/generate.py` rejects it, so the dictionary cannot go quiet as fields are added.
- **Ids are content-derived** — `sha256(path|unit|kind)` — so changing a field's unit or kind changes its id. A silent redefinition of an existing name is not expressible.

Episodic conditions are the case most likely to vary between implementations, so they are pinned in two parts: a `counter` totalling occurrences and `current` fields describing the most recent (`last_underflow_duration_ms`, `last_decode_error_pts_ms`, `last_decode_error_reason`). Both arrive in the same `getAll()` snapshot, so an occurrence and the counters around it are always mutually consistent — which is the property a test can actually pin down.

If any individual definition is ambiguous as written, that is worth flagging per field — the dictionary is the thing I most want picked apart before this goes GREEN.

Ulrond:
Correction to my comment above: I wrote "45 fields across `video_decoder`, `audio_decoder`, `audio_sink` and `clock`" and left out `video_sink`, which is where most of the presentation-side figures actually live — the repeat, freeze, late-drop and render-latency counters. The dictionary now stands at **49 fields across six elements** — `video_decoder`, `video_sink`, `audio_decoder`, `audio_sink`, `clock`, `drm` — at `dictionaryVersion` 1.2.

metrics/current/com/rdk/hal/metrics/Capabilities.aidl
```diff
@@ -0,0 +1,55 @@
...
+parcelable Capabilities
+{
+    /**
+     *  Opaque identity of this product's declared set. Stable while the
+     *  declaration is unchanged, different the moment anything in it changes.
+     *  A consumer cache-keys its resolved name map on this rather than diffing
+     *  field lists, and a bug report needs only this value to pin exactly what
+     *  the device was serving.
+     */
+    String schemaId;
```
ramkumarpattabiraman: Need to define the schemaId in hfp-metrics.yaml?

metrics/current/hfp-metrics.yaml
```diff
@@ -0,0 +1,816 @@
...
+              unit: none
+              kind: current
+              id: '0x26d18e085c02dfd4'
+              provider: Driver
+              description: >
+                The SoC's own error code for the same fault, carried through uninterpreted. For
+                vendor diagnosis; never branched on.
+
+
+          events:
```
ramkumarpattabiraman, line 180: There are no functions to register these events and get these events by client. Should those need to be added?

metrics/current/com/rdk/hal/metrics/IMetricsSource.aidl
```diff
@@ -0,0 +1,135 @@
+/*
```
ramkumarpattabiraman, line 1: **Architectural gaps by AI**

**Architectural gaps / observations**
**1.**	No schema-change notification (main gap). Capabilities.aidl and IMetricsManager.aidl say a consumer "re-reads only when schemaId changes," but IMetricsManagerEventListener only reports source add/remove. There is no push signal for a schemaId/capabilities change, so consumers must poll getCapabilities() to detect it. Consider an onSchemaChanged(String schemaId) callback.
**2.**	Register/enumerate race is unspecified. registerEventListener() does not state whether already-live sources are replayed as onSourceAdded. A consumer must therefore registerEventListener() then getSourcePaths() and reconcile/dedup, but the required ordering and dedup contract isn't documented — leaving a window to miss or double-count a source that appears between the two calls.
**3.**	getSource() malformed-path behavior undefined. The doc defines null only for "not live." Behavior for a syntactically invalid path (wrong segment count, undeclared element) is unspecified — is it null or EX_ILLEGAL_ARGUMENT? Worth stating explicitly.
**4.**	Read-method failure semantics are thin. getAll() / getFieldsByName() return boolean with no documented exception conditions (e.g., source removed mid-call, atomic latch failure). Only false is mentioned. Since the source can disappear (per session lifecycle), consumers need defined behavior for reads racing with removal.
**5.**	getField() out-parameter pattern is awkward but valid. The out long[] value length 1 array workaround is documented, but it means the caller must allocate/inspect a single-element array; getFieldsByName with one name already covers this. Not a bug — just redundant surface.
**6.**	Service-name convention deviation (minor). serviceName = "MetricsManager" (PascalCase) diverges from the repo's stated preference of lowercase/dotted service names (e.g., "boot", "sensor.motion"). Acceptable if it matches the module's existing convention, but flag for consistency.
**7.**	writable/setField vs read-cadence. MetricElementInfo.pollCadenceMs and freshness apply to reads; there is no declared contract for when a setField() write becomes visible on a subsequent getAll() (immediately, or bounded by cadence). Consider documenting write-visibility ordering.
**8.** 
Items 1 and 2 are the substantive architectural gaps; the rest are documentation/contract tightening.

metrics/current/com/rdk/hal/metrics/IMetricsSource.aidl
```diff
@@ -0,0 +1,135 @@
...
+     *
+     *  Names are the bare field names from `getFields()`. Unknown names are
+     *  omitted from the result rather than raising an error, so a newer consumer
+     *  degrades cleanly on an older product.
+     *
+     *  @param[in]  names  : bare field names to read, e.g. "frames_decoded".
+     *  @param[out] values : those that exist on this source.
+     *  @returns boolean : true on success.
+     */
+    boolean getFieldsByName(in String[] names, out MetricKVPair[] values);
```
ramkumarpattabiraman: If getFieldsByName with None/NULL passed in names will provide the same functionality as getAll, then getAll can be removed.

Ulrond:
@ramkumarpattabiraman — all four landed, two of them changed the interface. Detail is in the updated PR description.

- **`schemaId`** — removed. `MetricProfileInfo` already carries `interfaceVersion`/`schemaVersion` per layer and `MetricDomainInfo` carries `dictionaryVersion`, so the declaration already answered it, per layer rather than device-wide. `derived` and `instances` came out for the same reason.
- **Events** — no register/get functions, deliberately. The catalog is a declaration: it pins the moment a vendor must detect to the field movements it must produce, so the VTS can assert it. Two sentences offering consumers a listener that doesn't exist are gone.
- **`getAll()`** — dropped. `getFieldsByName(null)` and `getFieldsById(null)` read every field, so there's no third method to keep in step.
- **AI review** — 1 and 2 are resolved by the above: no `schemaId` means nothing to signal, and sources are now permanent (a decoder is hardware the vendor always has; idle is not absent), so `getSourcePaths()` is static and `IMetricsManagerEventListener` is removed entirely. 3 is now stated — `EX_ILLEGAL_ARGUMENT` for a malformed path, `null` for a well-formed one this platform doesn't serve. 7 wasn't the gap: no HAL in this repo documents write visibility, and where it matters they add an `onPropertyChanged` callback — the real miss was the `binder::Status::Exception::` prefix, now fixed. 6 I'd push back on, as PascalCase is the repo majority (`VideoDecoderManager`, `AudioSinkManager`, `AVClockManager`, `PlaneControl`). 5 is already answered in the file.

One consequence worth knowing, since it follows from permanent sources: counters now run from service start and a consumer takes its own baseline for whatever window it reports. That works for counters but not for maxima — a maximum over a window isn't the difference of two maxima — so `max_freeze_duration_ms` and `sync_max_abs_offset_ms` are now `writable` and accept a write of `0`.

Ulrond:
Correction to my comment above, on the events point — @ramkumarpattabiraman you were right to ask for register/get functions, and I now agree. Some of what the profile declares cannot be carried by a read at all: a burst inside one capture interval collapses to its newest member in the counters, and reading faster does not recover the rest.

`IMetricsSource` now has `registerEventListener()` / `unregisterEventListener()`, and each occurrence arrives as a `MetricsEvent` on `IMetricsSourceEventListener` at the instant it happened. The kinds an element raises come back in the catalog as `MetricElementInfo.events`, so a consumer knows what to expect before it registers. It sits on the source rather than the manager, matching `IVideoDecoder` and the other resource interfaces.

Two things it deliberately does not have. There is **no buffer** — no sequence number, no caller-held cursor, no retention, no drop-oldest accounting. The callback is the delivery, and the counters move whether or not anyone is listening, so a consumer that registers late or misses a call loses that occurrence's detail and never the count. An earlier revision of this PR carried a per-source ring; that was the expensive half for a vendor and stays removed.

And the payload uses a new `MetricEventValue` (bare name) rather than `MetricKVPair` — which also settles the conflict Copilot flagged on `MetricKVPair.aidl`. `MetricKVPair` promises a fully-qualified four-segment name because a metric value read from a source has one and needs it once it outlives the call. A payload value belongs to the occurrence rather than to a field, so it has no such path to give; `MetricsEvent` carries `sourcePath` once for the whole event instead.

What stands from my earlier comment: `schemaId`, `derived` and `instances` are still out, `getAll()` is still dropped, and the **manager-level** `IMetricsManagerEventListener` (source appear/disappear) is still removed — permanent sources have nothing to announce, and 19 of the 22 service-entry interfaces in this repo have no listener.

Ulrond:
Closing this PR — the central metrics service design is rejected.

A single `com.rdk.hal.metrics` module means one component answers for another component's figures. Every metric has to be routed from whoever measures it, through a service that holds no measurement of its own, to whoever reads it. That middle hop buys nothing and costs a source registry, a liveness contract for sources appearing and disappearing, and a second place a figure can be wrong.

The replacement is #783: every metric is served by the HAL that produces it. `videodecoder` answers for `av.video_decoder.*`, `videosink` for `av.video_sink.*`, and so on down. The envelope those six share lives in `common/`, so `drm` — which already carried this shape — becomes the template rather than the exception.

Nothing here is lost. All 51 figures defined in this PR's metrics profile carry over to #783 with their identifiers unchanged, because an identifier is derived from the field's composed path, unit and kind, and never depended on which file declared it.

Branch `feature/754-metrics-hal` is deleted. The diff stays readable on this PR.

PR #758, opened 2026-07-29
Add API to relase the acquired avbufferhelper instance

### LINKED ISSUE #755: Task: Add an API to release avbuffer helper instance
### Goal

1] `releaseAVBufferHelperInstance()` API should be added to release the acquired AVBufferHelper instance
2] Both `getAVBufferHelperInstance()` and  `releaseAVBufferHelperInstance()` API should be declared within an `extern "C"` block to prevent C++ name mangling, ensuring that dlysm() can reliably locate these symbols when library is loaded dynamically
3] A common predefined library name should be adopted by all vendors and this name should be explicitly documented in the aidl file. This ensures that the middleware can dynamically load the vendor library using a consistent identifier without requiring vendor specific handling

Items [2] and [3] are required to avoid build time dependency between middleware and the avbufferhelper library from vendor

### Notes (Optional)

_No response_

susanmary007:
Add API to relase the acquired avbufferhelper instance

### DIFF
--- a/avbuffer/0.2.0.0/include/avbufferhelper.h
+++ b/avbuffer/0.2.0.0/include/avbufferhelper.h
@@ -33,6 +33,12 @@ extern "C" {
 
 using namespace std;
 
+/**
+Note: All vendors shall provide the AV Buffer Helper implementation as libavbufferhelper.so. 
+The middleware will use this predefined library name for dynamic loading, 
+eliminating the need for vendor-specific handling and avoiding build-time dependencies.
+*/
+
 class IAVBufferHelper
 {
 public:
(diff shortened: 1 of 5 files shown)


avbuffer/0.2.0.0/include/avbufferhelper.h
```diff
@@ -33,6 +33,12 @@ extern "C" {
```
hari22yuva, line 33: need to push the changes on older version of avbufferhelper.h like avbuffer/0.1.0.0 or 0.2.0.0/avbufferhelper.h these old version this change will not available 

just push on avbuffer/current/avbufferhelper.h so that next version has avbuffer will have changes

PR #761, opened 2026-07-29
feat(planecontrol,videodecoder)!: the capture plane owns capture end to end (#753)

### LINKED ISSUE #753: Task: Extend planecontrol with a graphics capture interface (ICapture) — decoded-frame DMA-BUF capture for decode-to-texture
## Motivation

Graphics Player / decode-to-texture certification requires the application to import decoded video frames as GPU textures (EGL DMA-BUF import) while the decoder runs continuously. Today there is no HAL surface for this: the ports that can do it each expose a different vendor-private path (legacy `enable-texture` behaviours differ per SoC — one emits RGBA with no usable fd, others have no path at all), and middleware cannot state a portable contract.

## Proposal

Extend **`com.rdk.hal.planecontrol`** with a graphics capture interface, rather than adding a standalone capture module.

**Why planecontrol owns this.** Capture-to-texture is a routing decision about where a decoder's output goes — which is precisely what this module already owns. `planecontrol` carries `PlaneType { VIDEO, GRAPHICS }`, `SourcePlaneMapping` and `IPlaneControl.setVideoSourceDestinationPlaneMapping()`, and it already carries `IGraphicsFbProvider` / `IGraphicsFbProviderListener` for graphics framebuffers. Capture is the case where the destination is the application's texture rather than a display plane, so it belongs beside the routing it is a variant of.

There is house precedent: **`audiomixer` carries `IAudioCapture`** rather than audio capture living in a module of its own. Same shape — capture belongs to the component that owns the output path it taps.

`IVideoDecoder` is unchanged — the decoder does not know where its output goes. `IPlaneControl` routes it to a display plane; `ICapture` routes it to a DMA ring.

`OperationalMode.GRAPHICS_TEXTURE` already exists on `videodecoder` and is advertised through `IVideoDecoderManager.getSupportedOperationalModes()`. What is missing is the consumer-side ring and its lifecycle, which is what this adds.

## Proposed interface

Names are capture-prefixed where the module already defines the unprefixed name — `planecontrol` ships `Property.aidl`, `PropertyKVPair.aidl`, `PlaneCapabilities.aidl` and `GraphicsFbCapabilities.aidl` today, so a bare `Property.aidl` or `Capabilities.aidl` would collide.

```text
planecontrol/current/com/rdk/hal/planecontrol/
  ICaptureManager.aidl              — resource enumeration + capability queries
                                      (Option A only — see "Open for maintainers")
  ICapture.aidl                     — per-resource: Id, Capabilities, open()/close(), property, event listener
  ICaptureController.aidl           — per-session, returned by open(): start/stop/acquire/release/setProperty
  ICaptureControllerListener.aidl   — onRingReady, onFrameAvailable, onStateChanged
  ICaptureEventListener.aidl        — system and state-change events
  CaptureCapabilities.aidl          — per-resource capability parcelable
  CaptureProperty.aidl              — enum of capture property keys
  CapturePropertyKVPair.aidl        — { CaptureProperty, PropertyValue }
  VideoFrameView.aidl               — per-frame addressing parcelable
  CaptureErrorCode.aidl             — capture-specific error codes
planecontrol/current/hfp-planecontrol.yaml — gains the per-product capture capability declaration
```

Lifecycle: `getCapture(id)` → `open(decoderId, listener)` → `ICaptureController` → set ring properties → `start()` → `acquireLatestFrame()` / `releaseFrame(slot)` → `stop()` → `close(controller)`.

### `ICaptureManager`

```aidl
package com.rdk.hal.planecontrol;
import com.rdk.hal.planecontrol.ICapture;

@VintfStability
interface ICaptureManager
{
    const @utf8InCpp String serviceName = "CaptureManager";

    /** Platform list of Capture IDs. */
    ICapture.Id[] getCaptureIds();

    /** Gets a Capture interface, or null if the ID is invalid. */
    @nullable ICapture getCapture(in ICapture.Id captureId);
}
```

### `ICapture`

```aidl
package com.rdk.hal.planecontrol;
import com.rdk.hal.planecontrol.CaptureCapabilities;
import com.rdk.hal.planecontrol.ICaptureController;
import com.rdk.hal.planecontrol.ICaptureControllerListener;
import com.rdk.hal.planecontrol.ICaptureEventListener;
import com.rdk.hal.planecontrol.CaptureProperty;
import com.rdk.hal.planecontrol.CapturePropertyKVPair;
import com.rdk.hal.videodecoder.IVideoDecoder;
import com.rdk.hal.PropertyValue;
import com.rdk.hal.State;

@VintfStability
interface ICapture
{
    @VintfStability
    parcelable Id {
        const int UNDEFINED = -1;
        int value;
    }

    /** Capabilities for this Capture resource. Not allowed to change between calls. */
    CaptureCapabilities getCapabilities();

    @nullable PropertyValue getProperty(in CaptureProperty property);
    CapturePropertyKVPair[] getPropertyMulti(in CaptureProperty[] properties);

    /** Opens a capture session against an active video decoder.
     *
     *  Preconditions:
     *  - decoderId references an IVideoDecoder whose PROP_OPERATIONAL_MODE
     *    includes OperationalMode.GRAPHICS_TEXTURE before start().
     *  - This resource is in State::CLOSED.
     *
     *  Behaviour:
     *  - Transitions CLOSED -> OPENING -> READY.
     *  - The caller configures ring shape via ICaptureController.setProperty
     *    (PROP_SLOT_COUNT, PROP_SLOT_SIZE_BYTES, PROP_DRM_FOURCC,
     *    PROP_DRM_MODIFIER) before calling start().
     *
     *  @exception EX_ILLEGAL_ARGUMENT if decoderId is invalid.
     *  @exception EX_ILLEGAL_STATE if this resource is not CLOSED. */
    @nullable ICaptureController open(
        in IVideoDecoder.Id decoderId,
        in ICaptureControllerListener captureControllerListener);

    /** Closes the capture session. Releases the ring and any vendor-side
     *  wiring. The underlying IVideoDecoder is not stopped — only the
     *  capture binding is undone. */
    boolean close(in ICaptureController captureController);

    boolean registerEventListener(in ICaptureEventListener captureEventListener);
    boolean unregisterEventListener(in ICaptureEventListener captureEventListener);
}
```

### `ICaptureController`

```aidl
package com.rdk.hal.planecontrol;
import com.rdk.hal.planecontrol.CaptureProperty;
import com.rdk.hal.planecontrol.CapturePropertyKVPair;
import com.rdk.hal.planecontrol.VideoFrameView;
import com.rdk.hal.PropertyValue;

@VintfStability
interface ICaptureController
{
    /** Starts the capture session.
     *
     *  Reserves an N-slot ring of slotCount x slotSizeBytes from the platform's
     *  pre-allocated CMA video region — either one shared dma_buf with
     *  offset-addressed slots, or N per-slot dma_bufs, per the vendor surface
     *  model. Wires the decoder's GRAPHICS_TEXTURE output into the ring.
     *  Transitions READY -> STARTING -> STARTED and fires onRingReady().
     *
     *  @exception EX_ILLEGAL_STATE if not in READY.
     *  @exception EX_SERVICE_SPECIFIC (CaptureErrorCode.OUT_OF_MEMORY) on CMA exhaustion. */
    void start();

    /** Unwires the decoder from the ring and releases the ring.
     *  The decoder is left as-is. STARTED -> STOPPING -> READY. */
    void stop();

    /** Acquire the newest Ready slot, transitioning it to Locked atomically.
     *  The decoder will not write into a Locked slot.
     *  Returns null if no Ready slot exists — never blocks, and never
     *  re-delivers a frame already returned. */
    @nullable VideoFrameView acquireLatestFrame();

    /** Release a previously acquired slot. Locked -> Free.
     *  slot MUST be a VideoFrameView.slot value previously returned.
     *  Idempotent: stale, already-released or unknown slots return silently. */
    void releaseFrame(int slot);

    boolean setProperty(in CaptureProperty property, in PropertyValue propertyValue);
    boolean setPropertyMulti(in CapturePropertyKVPair[] properties);
}
```

### Listeners

```aidl
@VintfStability
oneway interface ICaptureControllerListener
{
    /** Fires once per session, after start() has wired the decoder into the
     *  ring and before any onFrameAvailable. ringFd is the shared dma_buf on
     *  single-ring platforms; null on per-surface platforms, where each slot's
     *  fd arrives per frame in VideoFrameView.planeFds. */
    void onRingReady(
        in @nullable ParcelFileDescriptor ringFd,
        in int[] planeStrides,          /* NV12: [Y_stride, UV_stride] */
        in int slotCount,
        in int slotSizeBytes);

    /** Optional push wake-up when a frame transitions to Ready.
     *  Implementations MAY coalesce. Consumers pulling at a known cadence
     *  can ignore it — the pull path is complete without it. */
    void onFrameAvailable();

    void onStateChanged(in com.rdk.hal.State previousState,
                        in com.rdk.hal.State currentState);
}

@VintfStability
oneway interface ICaptureEventListener
{
    /** Hardware or system errors not tied to a single frame — the bound
     *  decoder leaving GRAPHICS_TEXTURE unexpectedly, CMA exhaustion
     *  mid-session, IOMMU fault. */
    void onSystemError(in CaptureErrorCode error, in @utf8InCpp String detail);

    /** The bound IVideoDecoder was detached or closed. The session is
     *  implicitly stopped. */
    void onDecoderDetached();
}
```

### Parcelables

```aidl
@VintfStability @Backing(type="int")
enum CaptureProperty
{
    PROP_SLOT_COUNT      = 0,   /* int: N ring slots; default per HFP */
    PROP_SLOT_SIZE_BYTES = 1,   /* int: per-slot bytes incl. plane alignment */
    PROP_DRM_FOURCC      = 2,   /* int: DRM FOURCC; DRM_FORMAT_NV12 required */
    PROP_DRM_MODIFIER    = 3,   /* long: DRM_FORMAT_MOD_LINEAR required */
    PROP_WIDTH           = 4,
    PROP_HEIGHT          = 5,
}

@VintfStability
parcelable VideoFrameView
{
    int    slot;                      /* [0, slotCount) — pass to releaseFrame().
                                         Required because per-surface offsets are ~0
                                         and cannot disambiguate a slot. */
    ParcelFileDescriptor[] planeFds;  /* Single-ring: all equal the shared ring fd,
                                         stable across frames. Per-surface: this
                                         slot's fd(s), cycling through <= slotCount. */
    int[]  planeOffsets;              /* NV12: [Y_off, UV_off]; 0 is valid */
    int[]  planeStrides;
    int[]  planeLengths;
    int    width;
    int    height;
    int    drmFourcc;
    long   presentationTimeNs;
}

@VintfStability
parcelable CaptureCapabilities
{
    int    maxSlotCount;
    int    maxSlotSizeBytes;
    int[]  supportedFourCCs;
    long[] supportedModifiers;
}

@VintfStability @Backing(type="int")
enum CaptureErrorCode
{
    OUT_OF_MEMORY        = 1,   /* CMA reservation refused */
    DECODER_NOT_TEXTURE  = 2,   /* bound decoder not in GRAPHICS_TEXTURE */
    FORMAT_MISMATCH      = 3,   /* ring format differs from decoder output */
    DECODER_BUSY         = 4,   /* decoder already bound to another capture session */
    HARDWARE_FAULT       = 5,
}
```

## Buffer contract

Independent of the interface that carries it:

- NV12 linear, with truthful per-plane offsets — addressing the actual buffer layout, not assumed from the format.
- Importable via `EGL_EXT_image_dma_buf_import` without translation: `planeFds[N]`, `planeOffsets[N]` and `planeStrides[N]` feed `EGL_DMA_BUF_PLANE<N>_FD_EXT` / `_OFFSET_EXT` / `_PITCH_EXT` directly.
- Correctly refcounted, and each frame atomically complete before it becomes acquirable.
- Decode proceeds at full rate regardless of consumer acquire cadence, which may be sparse or delayed. The all-slots-locked policy (stall vs drop-newest) is HFP-declared.
- All DMA-BUFs released on transition to `NULL`.

## Per-product declaration

The vendor advertises `OperationalMode.GRAPHICS_TEXTURE` via `IVideoDecoderManager.getSupportedOperationalModes()` and declares it in `hfp-videodecoder.yaml`; the capture resource set and its per-resource capabilities are declared in the `planecontrol` HFP. Advertising the mode is necessary but not sufficient — a product that declares it must actually emit frames when the decoder is placed in `GRAPHICS_TEXTURE`.

Asked for an unsupported configuration, the implementation fails the bind with a clear error rather than silently falling back to plane output. A silent fallback is the failure mode that costs a certification cycle to find.

## Who calls it

Two shapes are viable and the HAL contract is the same in both — they differ only in who holds the binding per frame:

1. **Middleware relays.** Rialto Server binds `ICapture` and forwards each frame descriptor to the consumer over its own IPC.
2. **Consumer binds directly.** Middleware admits the session and selects the graphics destination; the consumer binds `ICapture` itself and pulls frames straight from the HAL.

The second halves the IPC hops on a per-frame path at 60 fps, and each hop costs an `SCM_RIGHTS` pass of a dma_buf descriptor. It also requires nothing new on the middleware client interface. What it has to answer is which decoder instance the consumer may capture from, and how a binding middleware does not hold is invalidated when the decoder is reclaimed under contention.

**This does not change the HAL contract** — it changes who the client is. Both are served by the same interface, and the decision sits above this repository.

## Conformance

L1-testable in isolation via the standard AIDL fixture: bind the capture service → `getCapture(id)` → `open(decoderId, listener)` → set ring properties → `start()` → expect `onRingReady` → push a known H.264 ES into the bound decoder → `acquireLatestFrame()` → verify pixel content at the returned per-plane offsets → `releaseFrame(slot)` → `stop()` → `close()`.

The assertions that matter: never re-delivers a frame already returned; returns null rather than blocking when nothing new exists; per-plane offsets address the actual layout; release is idempotent and tolerates out-of-range slots.

## Acceptance criteria

- `ICapture` and its parcelables added under `planecontrol/current/`, documentation following the standard HAL methodology.
- Capability / feature-profile declaration for ring limits (max slots, formats, modifiers) and concurrent capture count.
- L1/L2 test skeletons consistent with the other AV modules.

## Open for maintainers

**How `ICapture` is obtained — maintainers' choice.** This changes what a vendor publishes, so it should be settled before implementation starts. Both work; the difference is service topology, not capability.

The 2026-07-28 re-scope moved capture into `planecontrol` and dropped `ICaptureManager`, on the reasoning that a capture destination should be discovered the way a plane is. **The module placement stands.** The accessor shape is re-opened here because writing the concrete AIDL made the trade visible: the Manager → Resource → Controller triad is the shape the other AV modules already use, and dropping it spends that consistency to save a published service name. That is a maintainer call, not an author one.

**Option A — `ICaptureManager` singleton** (as drafted above)

```aidl
interface ICaptureManager {
    const @utf8InCpp String serviceName = "CaptureManager";
    ICapture.Id[] getCaptureIds();
    @nullable ICapture getCapture(in ICapture.Id captureId);
}
```

Mirrors the Manager → Resource → Controller shape used by `videodecoder`, `audiosink` and `videosink`. Costs a second published service name from the `planecontrol` module.

**Option B — accessor on `IPlaneControl`**

```aidl
// on the existing IPlaneControl interface
@nullable ICapture getCapture(in int planeResourceIndex,
                              in ICaptureEventListener captureEventListener);
```

Mirrors the module's own existing idiom — `getGraphicsFbProvider(planeResourceIndex, graphicsFbProviderListener)` returns `@nullable IGraphicsFbProvider`. One published service for the module, and capture is reached the same way the graphics FB provider already is. Resource enumeration would then come from the existing plane enumeration rather than a separate `getCaptureIds()`.

Everything else in this proposal is unchanged by the choice — `ICaptureController`, the listeners, the parcelables and the frame semantics are identical either way.

Also open: final method naming on `ICaptureController`, and whether format mismatch is detected at `start()` when the bound decoder's configured output format differs from `PROP_DRM_FOURCC` / `PROP_DRM_MODIFIER`.

---

_Re-scoped 2026-07-28: originally proposed as a standalone `com.rdk.hal.capture` module with its own Manager → Resource → Controller triad. Moved into `planecontrol` because capture is a routing destination for a decoder's output, which is what that module already owns — and because `audiomixer` already carries `IAudioCapture` on the same reasoning._

_Description replaced 2026-07-29 with the concrete AIDL surface. This description is the source of truth for the interface._

Ulrond:
Closes #753.

## What this adds

A decoded-frame capture surface on `com.rdk.hal.planecontrol`, so an application can import decoded video frames as GPU textures (EGL DMA-BUF import) while the decoder runs continuously. Each port does this today through a different vendor-private path, so middleware cannot state a portable contract.

Capture is declared, configured and bound entirely on the capture plane. `IVideoDecoder` carries nothing about it.

## Capture is a plane type

`PlaneType.CAPTURE` sits beside `VIDEO` and `GRAPHICS`. The type says where a plane's pixels come from and where they go:

| Type | Pixels come from | Pixels go to | Interface |
|---|---|---|---|
| `VIDEO` | A mapped video source | The display | `IPlaneControl` |
| `GRAPHICS` | The client, frame by frame | The display | `IGraphicsFbProvider` |
| `CAPTURE` | A mapped video source | The client, frame by frame | `ICapture` |

A capture plane runs opposite to a graphics plane: a graphics plane carries frames from the client to the display, a capture plane carries decoded frames from the pipeline to the client. It is never composited, so alpha, z-order, position and display latency have no meaning on it.

```aidl
// IPlaneControl
@nullable ICapture getCapture(in int planeResourceIndex, in ICaptureEventListener captureEventListener);
```

This mirrors `getGraphicsFbProvider()` exactly — one published service for the module, the capture resource enumerated by the existing plane enumeration. There is no `ICaptureManager` and no `ICapture.Id`.

## The mapping is the binding

`setVideoSourceDestinationPlaneMapping()` maps a source to a `CAPTURE` plane exactly as it maps one to a display plane, and that mapping is the whole of the binding:

```aidl
@nullable ICaptureController open(in ICaptureControllerListener captureControllerListener);
```

The source captured is whatever is mapped to the plane. A source is mapped to one plane at a time, which is what limits a decoder to a single capture session. Because the decoder is named in exactly one place, there is no second place for the two to disagree.

The vendor layer configures whatever the mapped source's decoder requires, over whatever internal path the platform provides.

## Interface surface

| File | Role |
|------|------|
| `ICapture.aidl` | Per-plane capture resource: `getCapabilities()`, `getState()`, property reads, `open()` / `close()` |
| `ICaptureController.aidl` | Per-session: `start()`, `stop()`, `acquireLatestFrame()`, `releaseFrame()`, property writes |
| `ICaptureControllerListener.aidl` | `onPoolReady()`, `onFrameAvailable()`, `onCaptureError()` |
| `ICaptureEventListener.aidl` | `onSystemError()`, `onSourceUnmapped()`, `onStateChanged()` |
| `CaptureCapabilities.aidl` | What a plane delivers and how its pool behaves |
| `CaptureProperty.aidl` | `BUFFER_COUNT`, `DRM_FOURCC`, `DRM_MODIFIER`, `WIDTH`, `HEIGHT` |
| `CapturePropertyKVPair.aidl` | Capture property key/value pair |
| `VideoBufferView.aidl` | Dma-Buf addressing of one pool buffer, delivered once per session |
| `VideoFrameView.aidl` | One frame: buffer index and presentation time |
| `CaptureErrorCode.aidl` | Capture error codes |
| `State.aidl` | planecontrol resource lifecycle states |

Lifecycle: `setVideoSourceDestinationPlaneMapping(source → capture plane)` → `getCapture(planeIndex, eventListener)` → `open(controllerListener)` → `setProperty(WIDTH/HEIGHT/BUFFER_COUNT)` → `start()` → `acquireLatestFrame(releaseIndex)` → `stop()` → `close(controller)`.

## Frame delivery

**Addressing is delivered once.** `onPoolReady(VideoBufferView[])` carries every pool buffer with the file descriptors, offsets, strides, lengths, size, format and modifier that address it. None of that changes during a session, so the client imports each buffer into an EGLImage on receipt.

**A frame is an index and a time.** `VideoFrameView` is `bufferIndex` + `presentationTimeNs`, so a frame costs an int and a long on the wire rather than a `ParcelFileDescriptor` array per frame at 60 Hz.

**Release and acquire are one call.** `acquireLatestFrame(releaseBufferIndex)` frees the previous buffer and takes the next in one round trip. `VideoFrameView.NO_BUFFER` on the first call. `releaseFrame()` remains for the last frame of a session.

**The frame returned is the one due for presentation.** Audio latency and AV-sync correction are applied by the vendor layer, so a client that draws on receipt is in sync without computing anything. Frames whose presentation time has passed are dropped; frames whose time has not come stay queued.

`planeFds[N]` / `planeOffsets[N]` / `planeStrides[N]` feed `EGL_DMA_BUF_PLANE<N>_FD_EXT` / `_OFFSET_EXT` / `_PITCH_EXT` directly — no translation. A client caching EGLImages must key on `bufferIndex`, never on the file descriptor alone: under a shared-Dma-Buf pool every buffer carries the same descriptor and an fd-keyed cache silently freezes the picture.

## What a plane declares, what a session configures

`CaptureCapabilities` — `supportedCodecs` (`H264` required), `supportedFourCCs` and `supportedModifiers` (`NV12` and `DRM_FORMAT_MOD_LINEAR` required), `maxFrameWidth` / `maxFrameHeight`, `resize`, `maxBufferCount`, `stallsWhenPoolExhausted`.

`CaptureProperty` — `WIDTH` and `HEIGHT` are written before `start()`. `BUFFER_COUNT` is optional; the vendor chooses the depth where a session leaves it unset, and reports what it reserved at `onPoolReady()`. `DRM_FOURCC` and `DRM_MODIFIER` are **read-only**: the hardware writing the memory determines the layout, so the plane states it and the client reads it to import correctly.

`resize` false means `WIDTH` / `HEIGHT` must equal what the mapped source decodes to, and `start()` fails with `RESOLUTION_MISMATCH` otherwise. Nothing is scaled, rotated, cropped, colour-converted or tone-mapped on this path — shape and colour belong to the consumer and may change on any frame.

## Errors and lifetime

`onCaptureError()` reports failures not tied to a single acquire. `CaptureErrorCode` covers `OUT_OF_MEMORY`, `SOURCE_NOT_MAPPED`, `CODEC_NOT_CAPTURABLE`, `HARDWARE_FAULT`, `RESOLUTION_MISMATCH`, `COLOR_CONVERSION_UNSUPPORTED`, `FORMAT_UNSUPPORTED` and `INVALID_CONFIGURATION`. Nothing falls back to plane output.

`plane_control.md` specifies startup in either order (frames produced before `start()` are discarded, and starting capture may cost a decode interruption while the vendor reconfigures), shutdown in either order, and that imported EGLImages do not survive `stop()`.

## videodecoder

Its net change from `develop` is the removal of `OperationalMode`, `IVideoDecoderManager.getSupportedOperationalModes()` and `Property.OPERATIONAL_MODE`. Where a decoder's frames go follows from how it is wired — mapped to a display plane, returned over `onFrameOutput()`, or mapped to a capture plane — so there is no mode to select. `video_decoder.md` carries the routing table.

## Product profile

`hfp-planecontrol.yaml` declares the DPI9 capture plane: FHD, NV12 linear, H264, `resize: false`, one plane. The interface stays general and the product declaration narrows it, so a later platform declares more without an interface change.

## Points for reviewers

1. **`onStateChanged` sits on `ICaptureEventListener`**, not the controller listener — `videodecoder` puts state changes on the event listener and frame callbacks on the controller listener.
2. **`CaptureErrorCode` is numbered compactly** (1–8) rather than preserving holes. The enum is new here and has never shipped.
3. **`resize` is a capability, not a property.** A plane either scales or it does not; that is a hardware fact, not a per-session choice.
4. **Buffer count is left to the vendor by default.** No Netflix-defined count is known, and the vendor knows its own memory region and reference-frame needs.
5. **`captureCapabilities` is declared in the HFP**, reachable through `ICapture.getCapabilities()` — `graphicsFbCapabilities` works the same way.

## Governance

New interface on a GREEN component, so reviewer sign-off in `planecontrol/metadata.yaml` and `videodecoder/metadata.yaml` is set to `recheck` for the 14+5 cycle.


planecontrol/current/com/rdk/hal/planecontrol/ICapture.aidl
```diff
@@ -0,0 +1,210 @@
...
+     * start and stop the session, and acquire and release frames. Controller related
+     * callbacks are made through the `ICaptureControllerListener` passed into the call.
+     *
+     * The client configures the ring shape through `ICaptureController.setProperty()` in
+     * the `READY` state, before calling `ICaptureController.start()`.
+     *
+     * The bound decoder must have `OperationalMode.GRAPHICS_TEXTURE` selected in its
+     * `Property.OPERATIONAL_MODE` before `ICaptureController.start()` is called. Whether
+     * that mode is available is advertised by
+     * `IVideoDecoderManager.getSupportedOperationalModes()`.
```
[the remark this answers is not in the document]
Ulrond: Fixed in ac45760d — qualified as `videodecoder.OperationalMode.GRAPHICS_TEXTURE`, `videodecoder.Property.OPERATIONAL_MODE` and `videodecoder.IVideoDecoderManager.getSupportedOperationalModes()` across all the capture AIDL and the module docs, not just this line.

planecontrol/current/docs/plane_control.md
```diff
@@ -344,6 +366,84 @@ SourcePlaneMapping[]=
 }
 ```
 
+## Decoded Frame Capture
+
+Capture routes a video decoder's output into a ring of Dma-Buf slots that the client imports as GPU textures, instead of routing it to the display plane. It is the case where the destination of a decoder's output is the application's texture rather than a plane, so it is reached through `IPlaneControl.getCapture()` on the video plane the decoder would otherwise have been mapped to.
+
+The `IVideoDecoder` contract is unchanged - the decoder does not know where its output goes.
+
+Capture requires the decoder to be operating in `OperationalMode.GRAPHICS_TEXTURE`, which is advertised by `IVideoDecoderManager.getSupportedOperationalModes()` and selected through the decoder's `Property.OPERATIONAL_MODE`. A video decoder can be bound to at most one capture session at a time.
```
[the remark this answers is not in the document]
Ulrond: Fixed in ac45760d — qualified as `videodecoder.OperationalMode.GRAPHICS_TEXTURE`, `videodecoder.Property.OPERATIONAL_MODE` and `videodecoder.IVideoDecoderManager.getSupportedOperationalModes()` across all the capture AIDL and the module docs, not just this line.

Ulrond:
## Requirements coverage check

Verified this PR against the graphics-player requirement set. Seventeen requirements: **twelve already met, one contradiction, four unstated.** All five are now addressed in `d5945252`.

### Already met

| Requirement | Where |
|---|---|
| Frames as Dma-Buf importable via `EGL_EXT_image_dma_buf_import` | `VideoFrameView.planeFds` |
| Per-plane offset and stride, truthful | `planeOffsets` / `planeStrides`, `HAL.PLANECONTROL.11` |
| NV12 + linear | Required of every decoder that supports capture |
| Producer-managed lifetime, lock and release | Free/Ready/Locked, `acquireLatestFrame()` / `releaseFrame()` |
| Pull surface primary | `acquireLatestFrame()`, never blocks, never re-delivers |
| Optional push wake-up | `onFrameAvailable()`, documented coalescable |
| Cleanup on every teardown path | `stop()` releases locked buffers, `close()` the pool, client crash implies both |
| Capability declaration matches reality | `HAL.VIDEODECODER.14/15` |
| Audio path conventional | Out of scope, correctly |

One deliberate deviation: the requirement says a consumer locks a frame **by offset**. We key by `bufferIndex`, because where the pool is not one shared allocation every buffer has an offset of 0 and an offset cannot identify it. Same guarantee, and it works under both allocation models.

### The contradiction — now fixed

`CaptureConfig.width` / `.height` were documented as *"the width in pixels of the captured frames"*, which invites a vendor to **scale output to them**. The specification requires the opposite: frames arrive at the resolution the stream decodes to, in source colorimetry, with no scaling, rotation, crop, colour conversion, tone-mapping or gamma adjustment.

This mattered because the wrong reading is the helpful-looking one. A vendor scaling to the configured size would pass every other check in this interface and produce frames the consumer cannot use — shape and colour belong to the consumer, are applied per frame, and can change on any frame, so a transform applied upstream would have to be undone.

Now stated in `CaptureConfig`, `VideoFrameView`, both module docs, and as **`HAL.PLANECONTROL.14`** and **`HAL.VIDEODECODER.16`**. The fields are restated as the maximum dimensions the buffers must accommodate; `VideoFrameView.width` / `.height` report what each frame actually is.

### The four unstated — now added

- **`HAL.PLANECONTROL.15`** — no more than one dropped frame per 15 seconds of capture, 144p through 2160p, while the client acquires and releases at cadence. A client that stops releasing is explicitly *not* covered; that case is `stallsWhenPoolExhausted`.
- **`HAL.PLANECONTROL.16`** — presentation time carried unaltered, and stated as the frame's only timing reference. A captured frame goes to the client's scene rather than a display plane, so the client presents it against the clock its audio path already runs on.
- **Concurrency** — capture planes and video planes are independent resources, so a product declaring both runs capture alongside a playback session routed to a display plane, up to the decoder count its video decoder profile declares.
- **Alpha** — no interface change needed. `supportedCaptureFourCCs` is an open list, so a product able to emit a format carrying alpha declares it and a client selects it. NV12 + linear is the required baseline, not the limit.

No interface signatures changed — this is documentation and requirements only, on top of the surface already under review.

planecontrol/current/com/rdk/hal/planecontrol/capture/VideoFrameView.aidl
```diff
@@ -0,0 +1,65 @@
...
+ *  as one `VideoBufferView` per pool buffer, so a client resolves a frame by looking up
+ *  `bufferIndex` in what it already holds.
+ *
+ *  @see VideoBufferView, ICaptureController.acquireLatestFrame()
+ *
+ *  @author    Gerald Weatherup
+ */
+
+@VintfStability
+parcelable VideoFrameView
```
ramkumarpattabiraman, line 38: Add DMA Buf file handle along with the offset, PTS.

planecontrol/current/com/rdk/hal/planecontrol/capture/VideoBufferView.aidl
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * A CLIENT CACHING EGLImages MUST KEY THE CACHE ON `bufferIndex`, or equivalently
+     * on the pair (file descriptor, offset) - NEVER on the file descriptor alone.
+     * Where the pool is one shared Dma-Buf, every buffer carries the SAME descriptor
+     * and differs only by offset, so a cache keyed on the descriptor collapses the
+     * whole pool onto one entry and the client re-textures a single buffer for the
+     * rest of the session. The picture freezes while frames continue to arrive, which
+     * is not a failure the client can see in what it was handed.
+     */
+    ParcelFileDescriptor[] planeFds;
```
ramkumarpattabiraman, line 91: Does this variable and the below ones after this all need to be an array? or just a single variable.
We already have the VideoBufferView as an array in void onPoolReady(in VideoBufferView[] buffers); call. So it would mean we are returning array inside an array.

We have one pool of capture buffer, either each buffer having its own unique fd + one buffer offset 0 or all buffers have same fd but different buffer offset.

planecontrol/current/com/rdk/hal/planecontrol/capture/VideoBufferView.aidl
```diff
@@ -0,0 +1,138 @@
...
+     * on, because naming a file descriptor across a binder boundary means passing
+     * a ParcelFileDescriptor back - a dup and an SCM_RIGHTS pass on every
+     * released frame - to name memory the implementation already has open. An
+     * index carries the same information in an int.
+     *
+     * The index is also the safer key across teardown. A stale index after a
+     * stop/start names nothing and is ignored; a stale file descriptor names
+     * memory that may since have been freed.
+     */
+    int bufferIndex;
```
ramkumarpattabiraman, line 67: Should this be called bufferOffset and not bufferIndex?

Ulrond: An index. A buffer's identity is its index and nothing else — the addressing in `planeFds`/`planeOffsets` is how a client reaches the memory, and this interface constrains it no further than having to address the buffer. An offset is therefore not promised to be distinct between buffers, and neither is a descriptor, so keying on either means keying on something the contract never guaranteed.

`1311c80c` rewrites this block to state that as the requirement, rather than explaining it by what implementations happen to do.

planecontrol/current/com/rdk/hal/planecontrol/capture/VideoBufferView.aidl
Ulrond: Per plane, not per buffer — the nesting is one level, not two.

`onPoolReady()` delivers one `VideoBufferView` per buffer, and that is the outer array. These arrays index the planes *within* one buffer, feeding `EGL_DMA_BUF_PLANE<N>_FD_EXT`, `_OFFSET_EXT` and `_PITCH_EXT` 1:1. `DRM_FORMAT_NV12` is required of every capture plane and has two planes, `[Y, UV]`, so length 2 is the baseline rather than an edge case; a single-plane format gives length 1. So it is buffers × planes.

The two arrangements you describe are exactly why the fields are carried per buffer: whether two buffers share a descriptor and differ by offset, or carry different descriptors at offset 0, a client imports from `planeFds` and `planeOffsets` alone and does not need to know which it was handed.

planecontrol/current/com/rdk/hal/planecontrol/capture/VideoFrameView.aidl
Ulrond: The descriptors are delivered — once, not per frame. `ICaptureControllerListener.onPoolReady()` carries one `VideoBufferView` per pool buffer with `planeFds`, `planeOffsets`, `planeStrides` and `planeLengths`. A client imports each buffer there, and thereafter needs only to be told which buffer holds the current frame, which is what `VideoFrameView` carries alongside the PTS.

Sending them per frame would not do the job even setting the cost aside: a `ParcelFileDescriptor` is duplicated as it crosses the binder boundary, so the same memory arrives as a **different integer on every delivery**. A per-frame descriptor cannot be compared against the previous frame's, so it cannot identify the buffer — which is the job `bufferIndex` does.

videodecoder/current/com/rdk/hal/videodecoder/Property.aidl
```diff
@@ -66,26 +66,6 @@ enum Property {
 	 */
     OUTPUT_FRAME_POOL_SIZE = 2,
```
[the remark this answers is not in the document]
Ulrond: Deliberate, and marked `!` on both the commit and the PR title. Two things make it a versioned removal rather than silent breakage:

- **The ordinal is retired, not reused.** `Property` now runs `… 2, 4, 5 …`, so no existing member's value shifts and no client misreads a value it still knows.
- **`current` is the development surface.** `videodecoder` is pre-baseline at `0.2.0.1`, and a client needing the old surface pins the frozen `videodecoder/0.1.0.0` or `0.2.0.0` snapshots, which this PR does not touch.

`OPERATIONAL_MODE` and `getSupportedOperationalModes()` move to the capture plane, which now owns capture end to end — that is the substance of #753 rather than a side effect of it. Capture is selected by mapping a source to a plane of type `CAPTURE`, so there is no longer a mode to set on the decoder.

videodecoder/current/com/rdk/hal/videodecoder/IVideoDecoderManager.aidl
```diff
@@ -50,24 +49,6 @@ interface IVideoDecoderManager
      */
 	IVideoDecoder.Id[] getVideoDecoderIds();
```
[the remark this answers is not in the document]
Ulrond: Same answer as on `Property.OPERATIONAL_MODE` — deliberate, marked `!`, and the capability moves to the capture plane, which owns capture end to end after #753. A client on the old surface pins the frozen `videodecoder/0.1.0.0` or `0.2.0.0` snapshots, which are untouched.

Ulrond:
## Finding — capture does not relay colour metadata it already has

Raising this separately because it is a pre-merge gap with no detectable failure mode, and it is the one thing on this PR I would not merge without a decision on.

**What is missing.** A capture plane specifies the memory layout of the frames it delivers and nothing about how to interpret their pixels. Not in `VideoBufferView`, not in `CaptureProperty`, not in `CaptureCapabilities`.

**Why it matters.** A consumer importing NV12 through `EGL_EXT_image_dma_buf_import` supplies the conversion parameters in the attribute list, and the external-texture sampler uses what it is given:

```
EGL_YUV_COLOR_SPACE_HINT_EXT               EGL_ITU_REC601_EXT / REC709_EXT / REC2020_EXT
EGL_SAMPLE_RANGE_HINT_EXT                  EGL_YUV_FULL_RANGE_EXT / EGL_YUV_NARROW_RANGE_EXT
EGL_YUV_CHROMA_HORIZONTAL_SITING_HINT_EXT  EGL_YUV_CHROMA_SITING_0_EXT / _0_5_EXT
EGL_YUV_CHROMA_VERTICAL_SITING_HINT_EXT    EGL_YUV_CHROMA_SITING_0_EXT / _0_5_EXT
```

Omit them and the driver falls back to its own default, typically BT.601 narrow. HD and UHD content is BT.709 or BT.2020, so the result is shifted colour and crushed or stretched levels. **Wrong colour renders a plausible picture, not an error** — no return code, no log line, and no application-side test can catch it, because the consumer has no reference to compare against. It surfaces as a subjective PQ complaint late in certification and is expensive to trace back to an unset EGL attribute.

`drmFourcc` does not answer it. `DRM_FORMAT_NV12` states the sample layout and says nothing about primaries, transfer or range; the same fourcc is correct for BT.601 SD and BT.2020 UHD alike.

**The type already exists one module across.** `videodecoder` defines `Colorimetry { range, matrix, transfer, primaries }`, CICP-valued, and the decoder already produces it per frame in `FrameMetadata.colorimetry`, with `Capabilities.supportedColorimetries` declaring what it can detect. The value is known on the producer side at the moment each frame is written; it is simply not relayed. `planecontrol` already imports `videodecoder` types where the concept is the decoder's — `CaptureCapabilities` imports `Codec`, `PlaneCapabilities` imports `PixelFormat` and `DynamicRange` — so this is the established pattern, not a new coupling.

`DynamicRange` does not close it: it is `SDR / HLG / HDR10 / HDR10_PLUS / DOLBY_VISION`, a classification of what a plane supports, and it cannot express BT.709 vs BT.2020 matrix or limited vs full range.

**What is genuinely new: chroma siting.** It appears nowhere in `videodecoder` or `planecontrol` — zero occurrences. Wrong siting is a half-sample chroma shift: subtler than a wrong matrix, equally undetectable from application code, visible on high-contrast colour edges.

### To decide

1. **Per frame or per pool.** The producer models colorimetry per frame (`FrameMetadata.colorimetry`). Relaying it once at `onPoolReady()` on `VideoBufferView` narrows that, and would pin the first value if colorimetry changes mid-session — an ad break or a profile switch, not only a source switch. Four enums is 16 bytes with no descriptors, so the cost argument that keeps addressing out of `VideoFrameView` does not apply here. I would put it on `VideoFrameView` and follow the producer, unless we can state why the narrowing is safe.
2. **What `UNKNOWN` means.** All four `Colorimetry` fields default to `UNKNOWN`. If the contract stops at "relay it", an implementation can relay `UNKNOWN` for everything, the client leaves the attribute unset, and we are back on the driver's BT.601 default — the exact bug. The contract needs a stated fallback.
3. **How chroma siting is represented.** A new enum in `videodecoder` beside the other colour types, so it serves any consumer of `FrameMetadata` and not only capture.

Ulrond:
Resolved in `6608521f`. Updating my comment above, which raised these as open questions — they are now decided and implemented.

**A capture plane was declared twice.** `PlaneCapabilities` and `CaptureCapabilities` both stated its formats — `videodecoder.PixelFormat` against DRM FOURCC + modifier — and both stated its size ceiling, `maxWidth`/`maxHeight` against `maxFrameWidth`/`maxFrameHeight` (that second pair I missed first time round). `PlaneType.CAPTURE` retired alpha, z-order, position, size and display latency and left the format fields standing, so on the face of the interface all of them still applied.

`CaptureCapabilities` now governs, because it is the only one of the two that can state the answer: an `EGL_EXT_image_dma_buf_import` client needs a FOURCC **and** a modifier, and `PixelFormat` has no modifier. On a capture plane `PlaneCapabilities` describes routing — `planeIndex`, `type`, `sourceTypes` — and every other field **shall be empty, zero or false**. Stated as an obligation rather than as "has no meaning": a field left to mean nothing gets populated anyway and read anyway, whereas zero is visibly unset and a VTS test can assert it. Both duplicate pairs then dissolve rather than needing a precedence rule, because only one side of each is ever populated.

**Neither declaration said what the pixels mean.** `NV12` states the sample layout and nothing about the primaries, transfer, matrix or range the samples were coded against — the same fourcc is correct for BT.601 SD and BT.2020 UHD alike. A consumer that omits the EGL conversion attributes gets the driver's default, typically BT.601 narrow, and renders a plausible picture in the wrong colour with no return code, no log line and nothing to compare against.

`videodecoder` already had the type and the decoder already produces it per frame, so:

- `CaptureCapabilities.supportedColorimetries[]` — the capability, read before `start()`.
- `VideoFrameView.colorimetry` — the value, **per frame**, because an ad break or a profile switch changes it without changing the source and a value fixed at `onPoolReady()` would pin the first one silently. Four enums, no file descriptors.
- Declaring the capability declares an obligation: a plane with a non-empty array populates every field on every frame, and an `UNKNOWN` from such a plane is a conformance failure rather than a value.
- An empty array means the plane cannot report colorimetry, and the fallback is stated once so every consumer assumes the same thing — BT.601 limited to 576 lines, BT.709 limited to 1080, BT.2020 NCL limited above.

**Chroma siting had no representation anywhere.** `ChromaSite` and `ChromaSitePosition` are new in `videodecoder`, carried on `FrameMetadata` and `VideoFrameView`. Kept out of `Colorimetry`, which maps field by field onto `GstVideoColorimetry` — GStreamer keeps `GstVideoChromaSite` as its own type for the same reason.

`planecontrol` and `videodecoder` both build clean.

Ulrond:
Withdrawing my previous comment. `6608521f` is reverted in `f7ca4723` — the branch is back to `1311c80c` and nothing from that proposal stands.

It changed `videodecoder`, adding `ChromaSite` and `ChromaSitePosition` types and a `chromaSite` field on `FrameMetadata`. That module is not in scope for this change and I should not have pushed it. `FrameMetadata.aidl` is now byte-identical to `develop` again, and this branch touches `videodecoder` only through the three commits that predate this — `4739763c`, `d5945252` and `6db6d866`.

The gap raised in https://github.com/rdkcentral/rdk-halif-aidl/pull/761#issuecomment-5296043701 is **still open**, and the questions in it stand:

1. Which parcelable governs a capture plane's declaration, given `PlaneCapabilities` and `CaptureCapabilities` both state its formats and its size ceiling with no stated precedence.
2. Whether colorimetry is carried per frame or once with the pool.
3. What an implementation owes when a colorimetry field is `UNKNOWN`.
4. Where chroma siting lives — and specifically whether it can be expressed without changing `videodecoder`, since siting has no representation in either module today.

Nothing further will be implemented against this until those are agreed.

Ulrond:
## Update — the interface has moved a long way since the last review

@ramkumarpattabiraman this is worth a fresh read rather than a diff; the PR description is rewritten to match.

**Capture and graphics are now subpackages.** `com.rdk.hal.planecontrol.capture` and `.graphics`, leaving `com.rdk.hal.planecontrol` for planes themselves — 9 files instead of 24, and a doc per package. Capture has never shipped (zero capture files in the frozen `0.1.0.0` and `0.2.0.0`), so the package move costs no client anything.

**Capture is reduced to a control surface.** `CaptureProperty` and `CapturePropertyKVPair` are deleted. One call replaces them:

```aidl
boolean setFormat(in FormatLayout format);   // one row of supportedFormats
```

`FormatLayout` pairs a fourcc with a modifier valid *for that fourcc*. Two independent lists implied the full cross-product was selectable, most of which no plane can deliver — a client would have found out at `start()`. Passing a declared pair back makes an invalid selection inexpressible rather than merely rejected.

**Nothing is mandated any more.** `supportedCodecs` and `supportedFormats` state what a product can deliver, and no value is required in either. A platform that cannot capture a codec or a format does not list it, and a mandate it cannot meet would not change that. `setFormat()` is consequently required before `start()`.

**Pool depth is the platform's.** `maxBufferCount` is gone. The vendor calibrates depth from the throughput its decode and memory path sustains; the client learns it by counting what `onPoolReady()` delivers, which is where the number already was.

**`State` is five values, not eight.** `UNKNOWN`, `OPENING` and `CLOSING` had no observable moment — `open()` and `close()` return synchronously, so nothing could ever see them. `CLOSED` is now zero, so a default-constructed value reads as something true. `STARTING` earns its place: `start()` returns before the pool exists, and `onPoolReady()` is what moves it to `STARTED`.

**Two things the interface never said, both of which a client would have hit:**

- **The listeners are `oneway`**, so `onPoolReady()` and `onGraphicsFbReleased()` arrive on a binder thread with no GL context current. An import must be handed to the thread that owns the context. `VideoBufferView` also holds `ParcelFileDescriptor`, which is move-only and delivered by const reference — a client duplicates the descriptors it intends to import from.
- **Descriptor lifetime across teardown.** The client's descriptors are duplicated across binder and an imported image takes a further reference, so the memory outlives `stop()`. What `stop()` ends is the *content* guarantee, not the memory. No copy is needed for safety — only to keep a frame beyond the period the buffer is held.

**Both docs now carry a worked example, and the examples are compiled.** A scratch harness extracts the fenced C++ from the docs and builds it against the generated headers. It has caught three real defects so far: `onPoolReady` declared `void` where the backend generates `Status`, `PropertyValue::make<>` used on the parcelable rather than its nested union, and `pool = buffers` which cannot compile at all against a move-only `ParcelFileDescriptor`.

Requirements are renumbered per package — `HAL.PLANECONTROL.1–6`, `…CAPTURE.1–11`, `…GRAPHICS.1–3` — and `videodecoder` is untouched by this PR.

Ulrond:
## Two vendor questions this interface depends on

Neither blocks review of the shape, but both decide whether it can be implemented as written.

### 1. Broadcom can do `(fd, offset)` — SDK26 checked

The per-SoC analysis behind this work concluded Broadcom's capture path exposes no Dma-Buf at all, based on Nexus **18.3** (`plane 0 fd = -1`, a CPU-mapped `NEXUS_Surface_Lock` pointer, RGBA downscale). That verdict is out of date. URSR SDK26 has:

```c
/* fd is cached until NEXUS_Surface_Destroy */
NEXUS_Error NEXUS_Surface_CreateDescriptor_driver(
    NEXUS_SurfaceHandle surface,
    int *fd            /* GEM export from pixelMemory */ );

typedef struct NEXUS_SurfaceMemoryProperties {
    NEXUS_MemoryBlockHandle pixelMemory;   /* block used for the pixel buffer */
    unsigned pixelMemoryOffset;            /* offset from the start of pixelMemory */
    ...

NEXUS_Error NEXUS_Platform_CreateMemoryBlockDescriptor_driver(
    NEXUS_MemoryBlockHandle block,
    unsigned blockOffset,   /* must be 4kb aligned */
    unsigned size,          /* must be 4kb aligned */
    int *fd );
```

Three things follow:

- **`(fd, offset)` per plane is the right shape for Broadcom**, not a problem for it. Several surfaces can share one `pixelMemory` at different offsets.
- **Offsets must be carried, never computed.** `pixelMemoryOffset` is whatever the allocator chose and block descriptors are 4 KB aligned, so a client deriving the chroma start from `stride × height` would miss.
- **"Export once, keep for the session" is supported** — *"fd is cached until `NEXUS_Surface_Destroy`"*.

Both APIs are marked *"API is subject to change"* and are `attr{local=true}`, so same-process only. Whether the vendor HAL sits where it can call them is an integration question worth confirming.

The interface already spans both shapes — AML/V4L2 gives per-plane fds at offset `0`, Broadcom gives shared memory at non-zero offsets — so no change follows from this. It removes a doubt rather than creating work.

### 2. Where does AV1 film grain sit relative to the capture tap?

DPI-9 makes Film Grain Synthesis mandatory for AV1. Grain is synthesised **after** decode and must not enter reference frames, so hardware typically applies it in the display pipeline, downstream of the decoded picture buffer.

A capture plane taps decoded frames. So:

- **tap before grain** — the client's texture is clean while the display shows grain; the captured picture differs visibly from the displayed one
- **tap after grain** — matches the display, but synthesis has to run into the capture pool as well, which is silicon cost that may not exist

`HAL.PLANECONTROL.CAPTURE.6` says frames are delivered applying no scaling, rotation, crop, colour conversion, tone-mapping or gamma adjustment. Grain is not in that list and arguably should not be — it is normative decode output per the AV1 spec, not a post-process the HAL adds. So the requirement does not answer it either way, and nothing else in `planecontrol` or `videodecoder` mentions grain.

This needs a vendor answer before AV1 capture is specified, and it sits alongside the AV1 decode-to-texture question already open against every SoC for 2026.1.

Ulrond:
An alternative to this PR is open as #782, against #781: the same capture interface as a module in its own right, `com.rdk.hal.capture`, rather than nested under `planecontrol`.

The AIDL types are the same in both — #782 carries them over verbatim with only the package rewritten. What differs is discovery and binding: #782 adds `ICaptureManager` and an `ICapture.Id`, and binds with `ICapture.open(CaptureSource, listener)` in place of `setVideoSourceDestinationPlaneMapping()`.

This PR also changes `planecontrol` and `videodecoder`; #782 touches only `capture/`, leaving both byte-identical to `develop`. They are therefore reviewable as alternatives rather than as overlapping changes.

Ulrond:
Closing — the approach in #753 is **rejected as a method**.

Decode-to-texture is being taken forward as **#781** instead: capture as a module in its own right (`com.rdk.hal.videocapture`), bound to one named video sink, implemented by **PR #782**.

This is the outcome #781 anticipated:

> #753 places the same interface inside `planecontrol`. The two are alternative designs for decode-to-texture; whichever is adopted, the other closes.

Closing this PR and deleting its branch. Review effort moves to PR #782.

PR #765, opened 2026-07-31
gh #755 Add API to relase the acquired avbufferhelper instance

susanmary007:
*  Add API to relase the acquired avbufferhelper instance
  *  Add changes to support dynamic library loading of avbufferhelper library

### DIFF
--- a/avbuffer/current/avbufferhelper.h
+++ b/avbuffer/current/avbufferhelper.h
@@ -33,6 +33,13 @@ extern "C" {
 
 using namespace std;
 
+/**
+ * @brief Vendors shall provide the AV Buffer Helper implementation as libavbufferhelper.so.
+ *
+ * The middleware uses this predefined library name for dynamic loading, eliminating the need for
+ * vendor-specific handling and avoiding build-time dependencies.
+ */
+
 class IAVBufferHelper
 {
 public:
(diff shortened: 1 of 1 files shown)


Ulrond:
@susanmary007 - the intent of #755 is right, but this cannot merge in its current form. Three items are genuinely blocking; the rest are cheap hygiene worth doing while the file is open.

## Blocking

**1. Moving `getAVBufferHelperInstance()` into `extern "C"` breaks the released ABI.**
`avbuffer/0.2.0.0/avbufferhelper.h` shipped that function with C++ linkage, so its exported symbol is mangled. Any vendor `libavbufferhelper.so` built against 0.2.0.0 exports the mangled name, and middleware doing `dlsym("getAVBufferHelperInstance")` against it will resolve to null. This is a breaking change to a vendor-facing contract.

At 0.x that may well be an acceptable break, but it has to be declared rather than slipped in: label the PR with the correct change class, and state in the PR body that vendors must rebuild. If a transition is preferred instead, export both symbols for one release.

**2. `releaseAVBufferHelperInstance(IAVBufferHelper* instance)` contradicts its own documentation.**
The comment describes a singleton with an internal reference count, in which case the parameter is redundant and unchecked. Either drop it to match the singleton model, or specify exactly what happens when a pointer that is not the singleton is passed. This is the new API being frozen into a vendor header - getting the shape right now avoids a second break later.

**3. The reference-count contract is unspecified.**
State whether the count is thread safe, and what an unbalanced `releaseAVBufferHelperInstance()` call does (release without a matching acquire, or release after the count already reached zero). Without this, vendor implementations will diverge.

## Worth fixing while the file is open

4. `IAVBufferHelper` declares no destructor. The header tells callers they must not delete the instance but nothing enforces it; a protected non-virtual destructor does.
5. There are now two `extern "C"` blocks in the file. The one near the top is empty and dead - consolidate.
6. `extern "C"` removes name mangling; it does not give a C ABI. The function still returns a pointer to an abstract class with a vtable, so the vendor library and the middleware must be ABI-compatible C++ builds. Say so explicitly, otherwise the guarantee reads as stronger than it is.
7. The header is not C-includable - it uses `class`, `<cstdint>` and `using namespace std;` - so the `#ifdef __cplusplus` guards are noise. The new `inline constexpr` also makes this C++17 or later; document that requirement.
8. `using namespace std;` at file scope in a public header pollutes every translation unit that includes it. Pre-existing, but this is a good moment to remove it.

## Minor

#755 item [3] asks for the library name to be documented "in the aidl file". There is no AIDL for avbuffer; it is a plain header. Worth correcting the issue text so it does not imply a missing interface surface.

The PR also has no labels and no milestone.

Ulrond (approved):
let's get it merged in. since  this is the one that's used.

Issue #767, opened 2026-08-03
Task: Update hfp for audio, video decoders to support W3C media capabilities


foumia:
### Goal

Update hfp for audio, video decoders to support W3C media capabilities
Existing attributes in hfp are not modified unless it is needed for new schema to support media capabilities

### Notes (Optional)
Mored documentation on hfp - .yamls and schema in 
https://github.com/rdkcentral/rdk-hpk-documentation/blob/develop/hfp-reference/hpk-hfp-configuration.md
_No response_

PR #768, opened 2026-08-03
Update audio video decoder hfps

foumia:
Update audio video decoder hfps to support w3c media capabilities
Existing attributes in hfp are not modified unless it is needed for new schema to support media capabilities

Task: https://github.com/rdkcentral/rdk-halif-aidl/issues/767

### DIFF
--- a/audiodecoder/current/hfp-audiodecoder.yaml
+++ b/audiodecoder/current/hfp-audiodecoder.yaml
@@ -22,28 +22,402 @@
 
 # HAL Feature Profile for audiodecoder
 
-audiodecoder:                   # Component object begins
-    interfaceVersion: current
-    IAudioDecoder:              # Resource list
-      - 0:                      # Resource object begins
-         supportedCodecs:       # Array of codec capabilities
-           - AAC_LC
-           - HE_AAC
-           - HE_AAC2
-           - DOLBY_AC3
-           - DOLBY_AC3_PLUS
-           - DOLBY_AC3_PLUS_JOC
-           - DOLBY_AC4
-           - X_HE_AAC
-         supportsSecure: true
-      - 1:                      # Resource object begins
-         supportedCodecs:       # Array of codec capabilities
-           - AAC_LC
-           - HE_AAC
-           - HE_AAC2
-           - DOLBY_AC3
-           - DOLBY_AC3_PLUS
-           - DOLBY_AC3_PLUS_JOC
-           - DOLBY_AC4
-           - X_HE_AAC
-         supportsSecure: true
+audiodecoder: # Component object begins
+  interfaceVersion: current
+  Capabilities:
+    - 0:
+      codecCapabilities:
+        - PCM:
+          profiles:
+            - BASE:
+              maxBitrateInBps: 1536000
+              maxChannels: 8
+              maxSampleRateInHz: 192000
+              maxBitDepth: 32
+
+        - AAC:
+          profiles:
+            - LC:
+              maxBitrateInBps: 576000
+              maxChannels: 8
+              maxSampleRateInHz: 96000
+              maxBitDepth: 24
+            - HE_V1:
+              maxBitrateInBps: 576000
+              maxChannels: 8
+              maxSampleRateInHz: 96000
+              maxBitDepth: 24
+            - HE_V2:
+              maxBitrateInBps: 576000
+              maxChannels: 8
+              maxSampleRateInHz: 96000
+              maxBitDepth: 24
+            - ELD:
+              maxBitrateInBps: 576000
+              maxChannels: 8
+              maxSampleRateInHz: 96000
+              maxBitDepth: 24
+            - X_HE:
+              maxBitrateInBps: 576000
+              maxChannels: 8
+              maxSampleRateInHz: 96000
+              maxBitDepth: 24
+
+        - MPEG_AUDIO:
+          profiles:
+            - LAYER_1:
+              maxBitrateInBps: 448000
+              maxChannels: 2
+              maxSampleRateInHz: 48000
+              maxBitDepth: 16
+            - LAYER_2:
+              maxBitrateInBps: 384000
+              maxChannels: 2
+              maxSampleRateInHz: 48000
+              maxBitDepth: 16
+
+        - MP3:
+          profiles:
+            - BASE:
+              maxBitrateInBps: 320000
+              maxChannels: 2
+              maxSampleRateInHz: 48000
+              maxBitDepth: 16
+
+        - ALAC:
+          profiles:
+            - BASE:
+              maxBitrateInBps: 4608000
+              maxChannels: 8
+              maxSampleRateInHz: 192000
+              maxBitDepth: 32
+
+        - SBC:
+          profiles:
+            - BASE:
+              maxBitrateInBps: 345000
+              maxChannels: 2
+              maxSampleRateInHz: 48000
+              maxBitDepth: 16
+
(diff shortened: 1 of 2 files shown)


Ulrond:
Thanks for picking this up — the direction is right, and splitting codec/profile with per-profile limits is what we need for W3C Media Capabilities alignment. A few things to resolve before this can land.

## Blocking: the audio HFP doesn't parse into the intended structure

The two files use different indentation conventions and the audio one nests incorrectly.

Video is correct — `profiles` is indented under the codec key:

```yaml
- MPEG2_VIDEO:
    profiles:
      - MPEG2_SIMPLE:
          maxLevel: MPEG2_LEVEL_HIGH
```

parses to `{MPEG2_VIDEO: {profiles: [...], dynamicRange: [...]}}`

Audio is not — `profiles` sits at the same level as the codec key:

```yaml
- PCM:
  profiles:
    - BASE:
      maxBitrateInBps: 1536000
```

parses to `{PCM: null, profiles: [{BASE: null, maxBitrateInBps: 1536000, ...}]}`

The codec name and the profile name both become null-valued siblings rather than parents. This affects every codec and every profile in `hfp-audiodecoder.yaml`.

Same issue on the resource index, in both files. On `develop`, resource `0` parses as `{0: {...}}`. On this branch it parses as `{0: null, codecCapabilities: [...], supportsSecure: true}` — the index is no longer the parent of the resource object. That's a regression against the existing structure.

Worth adding an HFP parse/schema check to CI so this class of error can't merge.

## `maxLevel` contradicts the frame dimensions

Resource 1 declares `H265_MAIN_10` with `maxLevel: H265_LEVEL_6_2` and `maxFrameWidth: 1920`. Level 6.2 permits 8192×4320@120. Both resources carry identical `maxLevel` values with only width/height differing, which suggests these aren't yet real vendor figures.

This matters specifically for Media Capabilities: a client calling `decodingInfo()` with a Level 6.2 codec string will get back "supported" and then fail at decode. Either the levels need to be the real ones, or the header needs to state explicitly that `maxFrameWidth`/`Height`/`FrameRate` clamp the declared level — and the clamping needs to be normative, not a comment.

## `dynamicRange` double-declares HDR

`H265_MAIN_10_HDR10` exists as a profile and `HDR10` appears in `dynamicRange`. Two sources of truth that can disagree. We should pick one — I'd suggest `dynamicRange` and drop HDR from the profile enum, but either way it needs to be one place.

Also worth confirming `DOLBY_VISION` under H264 and AV1 is real for target silicon rather than inherited from the template.

## Bitrates and audio limits look like placeholders

Every H264, VP9 and AV1 profile is 50 Mbps, identical across profiles and across both resources. The file header says to use `0` where the vendor doesn't publish a cap — most of these should probably be `0` until we have real numbers, otherwise the per-profile granularity carries no information.

In audio, AAC LC / HE_V1 / HE_V2 / ELD / X_HE all carry 576 kbps / 8ch / 96 kHz / 24-bit. HE-AAC v2 is parametric-stereo by definition so 8 channels looks wrong, and 96 kHz isn't a plausible xHE-AAC ceiling. These need checking against the codec specs.

## Duplication

Audio resource 0 and resource 1 are byte-identical — 17 codecs duplicated wholesale, taking the file from 1.7 KB to 11.6 KB. That will drift on the first edit. If resources share a capability set we should be able to declare it once and reference it, rather than copying.

## `Capabilities` vs `IAudioDecoder`

Renaming the resource list key from `IAudioDecoder` to a generic `Capabilities` loses which interface the resource list belongs to. If a component ever exposes more than one interface this becomes ambiguous. Was there a reason for the rename?

## `supportedDynamicRanges`

Per Copilot's comment — the resource no longer declares it but `Capabilities.aidl` still exposes the field. Related to the point below.

## Coordination with #149

The structures this PR introduces — `codecCapabilities`, per-profile limits, the resource-list layout — have no AIDL backing on `develop`. They're defined in #149 (`CodecSupport.aidl`, `Profile.aidl`, reworked `Capabilities.aidl`), which also touches both of these YAML files. So this is effectively a schema change expressed only in data.

Suggest we land #149 first with its scope frozen to exactly what's needed here, then rebase this onto it — at which point this becomes a small, pure-data PR. Anything here that needs new AIDL (per-codec `dynamicRange` placement, the resource-list construct) should be pulled back into #149 as a minimal addition.

## One forward-looking note

Resource 0 at 4K and resource 1 at HD is a static split. Is 1080p decoder 1's standalone hard limit, or its limit while decoder 0 is running 4K? If it's the latter, we're solving concurrency by hardcoding the pessimistic case and permanently giving away capability.

Either way, the header should state now that these per-resource maxima are **standalone** maxima, not guaranteed-concurrent ones — so that when a joint-constraint declaration lands alongside `Capabilities`, it isn't a breaking reinterpretation of what these numbers already meant.

Ulrond:
@foumia  we need to review this

Ulrond:
@foumia - I have pushed the Pykwalify schemas and a validator onto this branch as c985fbfe, since the schema and the restructure it defines are the same change:

- `videodecoder/current/hfp-videodecoder-schema.yaml`
- `audiodecoder/current/hfp-audiodecoder-schema.yaml`
- `scripts/validate_hfp.sh`

Schema content is verbatim from `rdk-hpk-documentation/hfp-reference/`, under the standard licence header. Naming them `hfp-*-schema.yaml` places them in the module input hash and in the `cp -r current/ <version>/` snapshot freeze, so a released snapshot carries the schema its HFP was written against.

```bash
python3 -m venv .hfpvenv && .hfpvenv/bin/pip install pykwalify
PYTHON=.hfpvenv/bin/python scripts/validate_hfp.sh videodecoder audiodecoder
```

Both files currently fail against it.

**1. Resource entries are not nested under the resource id.**

```
- Value 'None' is not a dict.  /videodecoder/Capabilities/0/0
- Key 'codecCapabilities' does not match any regex '^\d+$'
```

`- 0:` followed by sibling keys parses as `{0: None, codecCapabilities: ...}`, so the id maps to null and the capability keys land at the resource-list level. The schema wants them one level deeper:

```yaml
Capabilities:
  - 0:
      codecCapabilities:
```

The reference file in `hfp-reference/videodecoder/hfp-videodecoder.yaml` is nested this way and validates clean. Same defect in both files, both resources.

**2. `schemaVersion` is required and absent.** The reference file carries `schemaVersion: "0.1.0"`.

**3. `supportedOperationalModes` is not defined in the schema.** #791 removes it from the videodecoder HFP, so rebasing on that once it lands clears this one.

**4. Dynamic range is now declared per codec, but the AIDL still declares it per resource.**

`codecCapabilities[].dynamicRange` has no counterpart: `CodecCapabilities.aidl` carries `codec`, `profiles`, `maxFrameRate`, `maxFrameWidth`, `maxFrameHeight`, `supportedOutputPixelFormats` and no dynamic range, while `Capabilities.aidl` carries a flat `DynamicRange[] supportedDynamicRanges` that this PR removes from the HFP.

The HFP follows the API, so the API has to move first. Per-codec is the right model - `H265_MAIN_10_HDR10` already encodes dynamic range in `CodecProfile`, while `HLG`, `HDR10_PLUS` and `DOLBY_VISION` have no profile values at all, so today the same fact is expressed twice and neither expression is complete. That wants its own change: move `DynamicRange[]` into `CodecCapabilities`, drop `Capabilities.supportedDynamicRanges`, and take dynamic range out of `CodecProfile`.

Until that lands, this PR's `dynamicRange` key describes something the interface cannot express.

